Issue #21, opened 2016-11-02
Client.notify throws an error when releaseStage & notifyReleaseStages are set


sbycrosz:
I think it's because Array#contains is not a JS function [Link](https://github.com/bugsnag/bugsnag-react-native/blob/master/lib/Bugsnag.js#L135)

PR #22, opened 2016-11-02
Fix minor JS bugs

sbycrosz:
https://github.com/bugsnag/bugsnag-react-native/issues/21
& I changed `for .. in` (iterates keys) to `for .. of` (iterates values)


kattrali:
Ah, nice catch, @sbycrosz. Thank you!

PR #23, opened 2016-11-07
ES6 Cleanup

TheSisb:
Just some minor house keeping

### DIFF
--- a/lib/Bugsnag.js
+++ b/lib/Bugsnag.js
@@ -1,6 +1,5 @@
 import { NativeModules } from 'react-native';
 
-
 const NativeClient = NativeModules.BugsnagReactNative;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Bugsnag.js
```diff
@@ -37,13 +36,10 @@ export class Client {
...
-
-      ErrorUtils.setGlobalHandler(function(error, isFatal) {
-        if (bugsnag.config.autoNotify) {
-          bugsnag.notify(error, function(report) {
-            report.severity = 'error';
-          });
+      
+      ErrorUtils.setGlobalHandler((error, isFatal) => {
+        if (this.config.autoNotify) {
+          this.notify(error, report => report.severity = 'error');
```
wordofchristian: I have a slight quibble with this line here. The implicit return is potentially problematic since the `notify` function uses the return value of this callback for it's own internal logic. 

Perhaps we should change it to `this.notify(error, report => { report.severity = 'error'});` to avoid future bugs getting introduced. 

Other than that I agree with the rest of these.

TheSisb: Pardon my ignorance, how does adding some `{ }` make the function more safe?

wordofchristian: Curly braces with an arrow function removes the implicit return.

lib/Bugsnag.js
```diff
@@ -114,8 +111,8 @@ export class Client {
     let type = metadata['type'] || 'manual';
     delete metadata['type'];
     NativeClient.leaveBreadcrumb({
-      name: name,
-      type: type,
+      name,
+      type,
       metadata: typedMap(metadata)
```
kattrali: The `metadata` key is still here, I’m not sure if that was intended or not.

wordofchristian: I think this is because it can't be shortened like the others since it needs to go through the `typedMap` function.

kattrali:
Thanks again, @TheSisb!

PR #26, opened 2016-11-14
Fix default redbox error handling in development

wordofchristian:
Loading bugsnag currently breaks the default redbox behavior in development. This is because the react native `ErrorUtils` has been changed. The `_globalHandler` member is now private and has been replaced with a [proper getter](https://github.com/facebook/react-native/blob/3137ba993dbc3645832283ab683d18950c4246ed/packager/react-packager/src/Resolver/polyfills/error-guard.js#L33-L35).


GeoffreyPlitt:
Thanks! Let us know when this is on NPM

PR #29, opened 2016-11-23
Robust example application for using bugsnag with react native

wordofchristian:
## Demonstrates
- Extracting common configuration for multiple files and environments
- Identifying users
- Handling rejected promises
- Logging handled exceptions
- Using breadcrumbs
  - Navigation
  - Submitting forms
  - Network requests

Most of the interesting code will be in the `app/` folder. Everything else is pretty boilerplate code from react-native init, except for the `.eslintrc`

## Todo

- Restore [this native java module](https://github.com/bugsnag/bugsnag-react-native/pull/29/files#diff-a3dd8ce9e24bf8f257bd20a99e240058) from the old example.


wordofchristian:
@kattrali should we add the example directory to the `.npmignore`?

kattrali:
Isn't it already there?

wordofchristian:
@kattrali I brought back the module but for some reason it is inaccessible from javascript. Anything obvious that is wrong as far as you can tell? 
screenshot 2016-12-08 13 40 36

Issue #32, opened 2016-11-30
[Android] Crash when trying to leave non-string breadcrumbs


cooperka:
### Current behavior

`leaveBreadcrumb` on Android can't process non-string data types.

For example, each of the following lines:

```js
client.leaveBreadcrumb('some type', { foo: 1 });
client.leaveBreadcrumb('some type', { foo: true });
client.leaveBreadcrumb('some type', { foo: {} });
```

Leads to a crash with the errors:

```
TypeError: expected dynamic type `string', but had type `int64'
TypeError: expected dynamic type `string', but had type `boolean'
TypeError: expected dynamic type `string', but had type `object'
```

They work fine on iOS, and strings do work on Android (e.g. `{ foo: 'bar' }`). The crash itself happens in [`readStringMap`](https://github.com/bugsnag/bugsnag-react-native/blob/master/android/src/main/java/com/bugsnag/BugsnagReactNative.java#L115) trying to execute `pair.getString("value")`.

### Repro steps

You can repro successfully with the current example app, using the code above.

### Expected behavior

According to `allowedMapObjectTypes` in [lib/Bugsnag.js](https://github.com/bugsnag/bugsnag-react-native/blob/master/lib/Bugsnag.js#L237), `int` and `boolean` should be accepted, and `Object` should be recursively processed.

PR #33, opened 2016-11-30
Fix breadcrumb crash

cooperka:
### Changes

Fixes https://github.com/bugsnag/bugsnag-react-native/issues/32 (see issue for details).

The case statement I added is similar to the one lower down in `DiagnosticsCallback#readObjectMap`. It seems like it would be better to use [`ReadableMap#getType`](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableMap.java#L26) and switch on [`ReadableType`](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReadableType.java), but I didn't want to change that logic in this PR.

This will probably cause merge conflicts with https://github.com/bugsnag/bugsnag-react-native/pull/29 due to changes in the example app. I'll happily rebase on that PR if that's easier.

### Verification steps

Test with the Android example app:

1. Set the API key to a dashboard you have access to
1. Press "leave breadcrumb"
2. Press "crash"
3. Verify the details in your BugSnag dashboard

### DIFF
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -112,7 +112,20 @@ public class BugsnagReactNative extends ReactContextBaseJavaModule {
     while (iterator.hasNextKey()) {
         String key = iterator.nextKey();
         ReadableMap pair = map.getMap(key);
-        output.put(key, pair.getString("value"));
+        switch (pair.getString("type")) {
+            case "boolean":
+                output.put(key, String.valueOf(pair.getBoolean("value")));
+                break;
+            case "number":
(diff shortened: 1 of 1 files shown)


kattrali:
Thank you for the report and the fix, @cooperka.

kattrali:
Hm, I think it would be best to skip the changes to the example app for now, and I think this patch could go in before the new examples, which include much more detail for using the library.

cooperka:
@kattrali, agreed. I've rebased without the example app changes (previously 8c2764b).

kattrali:
Thanks! Looks good :)

Kerumen:
@kattrali Can you publish a npm version with this fix? Thanks!

PR #40, opened 2016-12-07
Add "promise" as a dependency

### LINKED ISSUE #37: Requiring unknown module "promise/setimmediate/rejection-tracking"
This error happens already when I call `client = new Client(BUGSNAG_API_KEY);` Not in __DEV__ mode though. I'm using v.1.3.1. 

It looks like the error happens in line 53 in lib/Bugsnag.js

genymotion_-_google_galaxy_nexus_-_4_3_-_api_18_-_720x1280__720x1280__320dpi__-_192_168_56_101

wordofchristian:
Adding this as an explicit dependency since we use it in our library.

This should hopefully resolve #37.


package.json
```diff
@@ -6,6 +6,7 @@
   "nativePackage": true,
   "dependencies": {
     "npmlog": "^2.0.4",
+    "promise": "^7.1.1",
```
kattrali: I don't think we should pin to more than the major version here, since we have a very small dependency and promise is likely to show up elsewhere in the dependency tree.

package.json
```diff
@@ -6,6 +6,7 @@
...
     "xcode": "^0.8.9"
```
kattrali: 😮❗️ We can actually remove this dependency now.

Issue #53, opened 2016-12-26
NativeClient.notifyBlocking is not a function


macnube:
Not sure if I messed something up with the installation but I did verify that both libz.tbd and libBugsnagReactNative.a are in my linked libraries in xCode. The client seems to initialize without errors. 

![image](https://cloud.githubusercontent.com/assets/7122725/21482528/dbac6cea-cb53-11e6-82f3-afb745dfe685.png)

After initializing the client in a singleton module I attempt to send a test error using client.notify and I catch the following error from the Promise:

TypeError: NativeClient.notifyBlocking is not a function
    at Client._callee$ (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:111239:14)
    at tryCatch (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34034:29)
    at Generator.invoke [as _invoke] (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34308:12)
    at Generator.prototype.(anonymous function) [as next] (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34067:13)
    at tryCatch (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34034:29)
    at invoke (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34110:12)
    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:34155:1
    at tryCallTwo (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:33674:1)
    at doResolve (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:33829:9)
    at new Promise (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false:33695:1)

Crash--:
Seems this is a regression on 1.2.0 introduced by this commit https://github.com/bugsnag/bugsnag-react-native/commit/91db6339c464498068befc46ab4aa164a71163a3 since `notifyBlocking` was only added on Android :/

the-simian:
Yes i can also confirm that Bugsnag is cucrrently broken on 1.2.0 on ios, and this is the problem.

the-simian:
references https://github.com/bugsnag/bugsnag-react-native/pull/54

kattrali:
Fixed in c172297

PR #54, opened 2016-12-27
call notify if notifyBlocking is not defined

evollu:
I don't know if this is right fix for #53

### DIFF
--- a/lib/Bugsnag.js
+++ b/lib/Bugsnag.js
@@ -91,9 +91,13 @@ export class Client {
       beforeSendCallback(report);
     }
 
-    NativeClient.notifyBlocking(report.toJSON(),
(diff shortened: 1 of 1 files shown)


the-simian:
@evollu did you test this yet? I'll try to take a look if you haven't

evollu:
I'm using it now. seems fine. but IOS is using the none blocking notify

kattrali:
Thank you for the contribution! iOS does not require two different methods since it writes to disk synchronously in the case that the application crashes, as opposed to the Android which immediately sends the request. This is the ideal solution.

evollu:
cool good to know that

the-simian:
Thanks ya'll

PR #58, opened 2017-01-03
Provide a means of disabling automatic uncaught exception handling [Android]

mariostallone:
Use the `autoNotify` config option to disable automatic uncaught exception handling on Android.
I chose to use `autoNotify` as it would be simpler to use one key across android and iOS from React Native.
Please correct me, if these are two separate keys. I'll separate this into another key in that case.

The motivation behind this.

- A few of us use bugnsag primarily to track React Native induced exceptions
- This will prevent tracking of native exceptions on bugsnag, if one chooses not to


kattrali:
Thank you for the contribution, @mariostallone. 

It may be worth noting that a native exception can also be caused from JavaScript code, in the case that code is called from across the native bridge. A common example would be libraries for custom views or accessing hardware. With `autoNotify` disabled, this variety of error would not be reported since the native code would crash the app and the JavaScript error reporter would not be called. 

Would this be expected behavior, and is discarding those errors desirable?

mariostallone:
Thanks @kattrali.
Yes, I do understand that such exceptions wouldn't bubble down to bugsnag. 
This flag would primarily benefit those of us who are already using a separate native crash reporting tool. This would prevent clouding the logs on bugsnag, as we primarily use bugsnag to track react native exceptions.

kattrali:
Gotcha, thanks for the clarification.

mariostallone:
Hey @kattrali Just noticed that this hasn't been pulled in yet. Are there any concerns with the PR?

kattrali:
Sorry about that, @mariostallone. The option itself is fine, but naming is hard. Since the `autoNotify` option already exists, it would be an unexpected change to start disabling native reporting now if previously it was just used to toggle unhandled exception reporting from JS. So potentially `autoNotifyNative`? And then I'll need to port it to iOS as well, to keep the same interface between them. If you don't have the time to revisit it, its something I can take care of in a few days.

mariostallone:
This has already been resolved by @kattrali  in v2.2.2

PR #60, opened 2017-01-10
Update iOS header imports for RN 0.40

henrikra:
As you can see from [RN 0.40 release notes](https://github.com/facebook/react-native/releases/tag/v0.40.0) they have moved ALL native modules on iOS. All packages that use `.h`-files starting with `RCT` they have to be refactored to the new path.

So basically `bugsnag-react-native` won't work with RN 0.40 without this PR! :)

### DIFF
--- a/cocoa/BugsnagReactNative.h
+++ b/cocoa/BugsnagReactNative.h
@@ -1,6 +1,6 @@
 #import <Foundation/Foundation.h>
 
-#import "RCTBridgeModule.h"
+#import <React/RCTBridgeModule.h>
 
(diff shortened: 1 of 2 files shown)


wordofchristian:
This actually breaks the example project. It would need to be upgraded to react-native 0.40 by following this procedure https://facebook.github.io/react-native/docs/upgrading.html

I think we'd rather get this out sooner than later, so it might be preferable to revert the changes in the `example/` directory for now.

henrikra:
I updated the PR. Anything else you want to change? :)

wordofchristian:
Looks good, thanks @henrikra. We'll release a new version momentarily

PR #61, opened 2017-01-10
Add original error to report object

dingbat:
Sometimes the error object itself contains some important application context that might need to be reported to Bugsnag. This allows the original error object reported to be available in the report, perhaps to be added to metadata in `beforeSendCallbacks`, for example:

```
configuration.beforeSendCallbacks.push(report => {
    const error = report.error;
    ...
})
```

I thought this was a fairly simple solution, but let me know if there's a better way of doing this.

### DIFF
--- a/lib/Bugsnag.js
+++ b/lib/Bugsnag.js
@@ -224,6 +224,7 @@ export class Report {
     this.severity = 'warning';
     this.stacktrace = error.stack;
     this.user = {};
+    this.error = error;
   }
 
   /**


kattrali:
Thank you for the contribution, @dingbat. This makes a lot of sense. My only concern would be the assumption that if the error is on the report, it’ll be propagated to the request sent to the server which is not the case. Perhaps it could be a second argument passed to the callback block. I’m open to some ideas here.

cooperka:
Just a thought, what about using a single object to store all the report data that's being sent to the backend, and a second var to store this new `error` object? Then `toJSON()` can simply `return reportData`, and it's clear that `error` is separate from that.

e.g.

```js
  export class Report {

    constructor(apiKey, error) {
      this.reportData = {
        apiKey,
        errorClass: error.constructor.name,
        errorMessage: error.message,
        context: undefined,
        groupingHash: undefined,
        metadata: {},
        severity: 'warning',
        stacktrace: error.stack,
        user: {},
      }
      this.originalError = error;
    }

    toJSON = () => {
      return {
        ...this.reportData,
        metadata: typedMap(this.reportData.metadata),
      };
    }
  }
```

kattrali:
@cooperka That's a interesting take on this as well. Would this provide an advantage to passing the error to the callback block?

cooperka:
I think the only real advantage with my suggestion is code cleanliness and readability; it makes it a lot more obvious whats going on.

If you're serializing a set of data to send to the backend, it makes sense to me that it should be stored in a map rather than a series of variables. Then anything that's not being sent is clearly distinguished, and any changes you're making to the data (such as transforming `metadata`) are also very clear.

As you suggested, adding the error as a second param might also make sense. The callback could be passed the values `(report.reportData, report.originalError)` instead of simply `(report)`. I like that a lot.

kattrali:
I opted to add the error parameter to the callback block, where it will be available to the user. There will be a release by the end of the week. Thanks again!

PR #64, opened 2017-01-17
turn off auto notify in debug mode

evollu:
just like handlePromiseRejections, prefer redbox in dev mode so it doesn't send the bug to bugsnag

### DIFF
--- a/lib/Bugsnag.js
+++ b/lib/Bugsnag.js
@@ -154,7 +154,7 @@ export class Configuration {
     this.notifyReleaseStages = undefined;
     this.releaseStage = undefined;
     this.appVersion = undefined;
-    this.autoNotify = true;
+    this.autoNotify = !__DEV__;
     this.handlePromiseRejections = !__DEV__; // prefer banner in dev mode
   }
 


kattrali:
Thanks for the contribution, @evollu. The current behavior in debug mode should be to do both -- notify bugsnag and show the red box. Is that not happening? In general, `autoNotify` being on by default avoids confusion during initial setup where a user may not understand why errors are not immediately being caught, though the option is easily changed.

evollu:
It is just annoying to receive error reporting email in while developing. Overriding default config is easy so you can close this if this is expected behavior

wordofchristian:
@evollu It is recommended that after confirming that the bugsnag is working in development, to change [this setting](https://docs.bugsnag.com/platforms/react-native/configuration-options/#notifyreleasestages) so you won't get any more development notifications.

```js
configuration.notifyReleaseStages = ['beta', 'production'];
```

kattrali:
Ah, thanks for the clarification @evollu. This behavior is intended but we can revisit if it becomes a usage problem.

@wordofchristian Thanks, that's also a good point. Maybe its something we can clarify in the initial usage guide.

Closing for now.

PR #67, opened 2017-01-26
fix: call global error handler regardless of bugsnag reporter

sondremare:
In the case where one configures releaseStage and notifyReleaseStages and is running a stage that is configured not to report to bugsnag, react native's global error handler is never called. This will make sure that any global error handler will be called regardless of what "shouldNotify()" returns.

### DIFF
--- a/lib/Bugsnag.js
+++ b/lib/Bugsnag.js
@@ -39,12 +39,9 @@ export class Client {
 
       ErrorUtils.setGlobalHandler((error, isFatal) => {
         if (this.config.autoNotify) {
-          this.notify(error, report => {report.severity = 'error'}, !!NativeClient.notifyBlocking, () => {
-            if (previousHandler) {
-              previousHandler(error, isFatal);
-            }
-          });
-        } else if (previousHandler) {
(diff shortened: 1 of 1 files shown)


kattrali:
Thanks for the report and the fix, @sondremare. Looking it over now.

lib/Bugsnag.js
```diff
@@ -39,12 +39,9 @@ export class Client {
 
       ErrorUtils.setGlobalHandler((error, isFatal) => {
         if (this.config.autoNotify) {
-          this.notify(error, report => {report.severity = 'error'}, !!NativeClient.notifyBlocking, () => {
-            if (previousHandler) {
-              previousHandler(error, isFatal);
-            }
-          });
-        } else if (previousHandler) {
+            this.notify(error, report => {report.severity = 'error'}, !!NativeClient.notifyBlocking);
```
kattrali: Unfortunately, we can’t just remove the callback block here, as its needed to ensure android reports are written in the event of a native crash. Perhaps a better solution would be to change the initial call check to be 

```js
if (this.config.autoNotify && this.config.shouldNotify())
```

so it will fall through in the event of release stage filtering?

sondremare: I will get on it

PR #81, opened 2017-03-28
fix: send uncaught errors to the original error handler when one should not notify Bugsnag

sondremare:
Added a check for shouldNotify before notifying Bugsnag of any uncaught error. This error was found using releaseStages, and running the app with a releaseStage that should not notify. Since it should not notify, the callback was never called. And with that, the previousHandler was never called.


kattrali:
Looks good, thanks @sondremare!

PR #89, opened 2017-04-13
Add warning when breadcrumb name exceeds max length

cooperka:
Anecdotally, it appears that breadcrumb names get truncated at 30 characters. I didn't see this specifically mentioned in the docs, but a few of my events were truncated online so it seems like it would be good to have a warning for this.


kattrali:
Thanks!

PR #101, opened 2017-05-04
Don't remove type property with mutation

dejakob:
You should not remove the type property of metadata. It will mutate the original object and can cause some issues in the project which uses this module. It's better to create a new Object and remove the type property there.


kattrali:
Thanks, @dejakob!

PR #121, opened 2017-07-24
Ensure RCTFatalExceptions are skipped

### LINKED ISSUE #115: RCTFatalExceptions are reported when using CodePush
From @andriantosg (split from #35):

Hi, does anyone know why I still got `RCTFatalException: Unhandled JS Exception:` while using CodePush? I've tried using `*main.jsbundle` on minified-url but unfortunately It didn't work. Am I missing something?

kattrali:
In the case that JS initialization never happens, move up the essential elements of the JS config to ensure `RCTFatalException` messages are missed. 

Fixes #115 

Steps to test:

* Initialize a React Native app using the docs instructions, skipping the JS portion
* Configure bugsnag
* Run the app on iOS
* Ensure that causing a JS error does not generate an error report with the title `RCTFatalException ...`


kattrali:
Closing this in favor of a docs change better explaining the root cause.

PR #123, opened 2017-08-02
Add test harness

fractalwrench:
Creates an app which launches and immediately notifies a local fake server of an exception.


cocoa/BugsnagReactNative.xcodeproj/xcuserdata/jamielynch.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kattrali: Looks like we need to add `xcuserdata` to the gitignore file - everything in there is machine-/user-specific shouldn't be in the project

PR #141, opened 2017-09-14
Track handled unhandled

fractalwrench:
I think the Javascript side of things is ready for review.

There will need to be some changes to the BugsnagReactNative module for Android/Cocoa. Specifically, the added fields in the report payload will need to be added to the native code's payload. I thought it would make sense to wait until we're happy with the Android/Cocoa implementation before proceeding any further on this.


src/Bugsnag.js
```diff
@@ -269,7 +293,10 @@ export class Report {
       metadata: typedMap(this.metadata),
       severity: this.severity,
       stacktrace: this.stacktrace,
-      user: this.user
+      user: this.user,
+      defaultSeverity: this._eventHandledState.defaultSeverity === this.severity,
```
snmaynard: what sets defaultSeverity on _eventHandledState? Should this be originalSeverity?

src/Bugsnag.js
```diff
@@ -227,7 +230,15 @@ export class Configuration {
 export class StandardDelivery {
 
   constructor(endpoint) {
-    this.endpoint = endpoint || 'https://notify.bugsnag.com';
+    this.endpoint = endpoint || 'https://fake.bugsnag.com'; // FIXME
```
snmaynard: This will need to be changed back I assume :)

PR #143, opened 2017-09-14
Add initial set of unit tests to the js client

bengourley:
Added some tests for the js client. Used [jest](https://github.com/facebook/jest) as that seems to be what the React/ReactNative community use.

Managed to cover ~75% of the codebase. Would be good to get it up to 100% but 75% is better than 0 😆.

```
Test Suites: 1 passed, 1 total
Tests:       8 passed, 8 total
Snapshots:   0 total
Time:        0.913s, estimated 1s
Ran all test suites matching /src\/__tests__/i.
------------|----------|----------|----------|----------|----------------|
File        |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
------------|----------|----------|----------|----------|----------------|
All files   |    79.31 |    67.14 |    60.87 |     80.7 |                |
 Bugsnag.js |    79.31 |    67.14 |    60.87 |     80.7 |... 203,211,300 |
------------|----------|----------|----------|----------|----------------|
```


bengourley:
One thing to check. I moved the babel config into its own file so that it could be shared by the tests and the `prepublish` command. Should make sure the output from `prebulish` is still 🆗.

src/__tests__/Bugsnag.test.js
```diff
@@ -0,0 +1,190 @@
...
+
+  const { Client, Configuration } = require('../Bugsnag')
+  const config = new Configuration('API_KEY')
+  config.handlePromiseRejections = true
+  const c = new Client(config)
+
+  c.notify = jest.fn()
+  Promise.reject(new Error('pboom!'))
+  // promise rejection handler has a slight async delay before running, so check after minimal timeout
+  setTimeout(() => expect(c.notify).toHaveBeenCalledWith(expect.any(Error)))
```
bengourley: I think this actually just causes the assertion to happen after the test has finished running. Jest tests appear to be synchronous unless you return a promise 🤔

PR #153, opened 2017-10-04
Upgrade cocoa deps, add new instructions

kattrali:
* Updates bugsnag-cocoa version
* Add releasing instructions for updating vendored lib
* Resolves issue where callback severity reason is being collected as handled


CONTRIBUTING.md
```diff
@@ -39,7 +39,18 @@ to run the example projects.
...
+
+To upgrade bugsnag-cocoa, prepare an updated version of bugsnag-cocoa, then copy
+the Source and iOS directories. Assuming bugsnag-cocoa is located in the same
+directory as bugsnag-react-native:
+
+```
+cp -r ../bugsnag-cocoa/Source/* cocoa/vendor/bugsnag-cocoa/Source
+cp -r ../bugsnag-cocoa/iOS/* cocoa/vendor/bugsnag-cocoa/iOS
+```
+
```
fractalwrench: The iOS xcodeproj will also need to be copied over (especially if we're adding/removing source files)

cocoa/vendor/bugsnag-cocoa/Source/BugsnagHandledState.m
```diff
@@ -116,8 +120,9 @@ + (NSString *)stringFromSeverityReason:(SeverityReasonType)severityReason {
             return kPromiseRejection;
         case UserSpecifiedSeverity:
             return kUserSpecifiedSeverity;
+        case LogMessage:
+            return kLogGenerated;
         case UnhandledException:
-        default:
             return kUnhandledException;
```
fractalwrench: Should we use `NSParameterAssert` for the default case here?

kattrali: If there's no default case, there's a build warning if any cases are missing. The change here was to specify every case so its not needed.

CONTRIBUTING.md
kattrali: Isn't that in `iOS/`?

fractalwrench: Yes - my bad!

kattrali: No worries, I had to check!

PR #156, opened 2017-10-06
Use existing android configuration if available

### LINKED ISSUE #25: Bugsnag Android init Error
Hello

I am getting the following error running bugsnag in an Android (Genymotion) emulator.

You must call Bugsnag.init before any other Bugsnag methods.

error

This seems to only be happening with Hot module loading enabled but I can't be sure.

From what I can tell from the android library the client should never be null?

Thanks
Ciaran

kattrali:
Uses a previously-instantiated Bugsnag client if one is available rather than creating a new client for the react native context. Allows users to configure settings which will be persisted in the native layer and reduces chaining native exception handlers if more than one client exists and is configured to automatically notify. (May also affect hot reloading when configuration occurs in the JS layer)

Fixes #25


android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -175,6 +174,23 @@ public void clearUser() {
...
   }
 
+  private Client getClient(String apiKey) {
+      Client client = null;
+      try {
+          client = Bugsnag.getClient();
+      } catch (IllegalStateException exception) {
+          // Create a new client as none exists yet
+      }
+      if (client == null) {
```
fractalwrench: we can get rid of the null check by moving this into the catch block

kattrali: Ah true. Totally overlooked that `getClient` is `@Nonnull`

Issue #152, opened 2017-10-04
TVOS support


ndillon1:
Will this library support Apple TV?

fractalwrench:
![](https://github.trello.services/images/mini-trello-icon.png) [TVOS support](https://trello.com/c/7OaUDxP9/6-tvos-support)

fractalwrench:
@ndillon1 the current release of the library doesn't support tvOS. However, we're experimenting with support in a PR which I've referenced in this issue.

benoitdion:
👍 for tvOS support

ChrisCodeCole:
I was able to get this running on tvOS simply by changing the podspec in node_modules. The issue was MessageUI framework which is only available on iOS. Bugsnag files look the same between iOS and tvOS from a quick glance, but I haven't fully tested out the capabilities. We'll be using this version of Bugsnag and I'll update if we run into any issues. My changes were:

- added **s.tvos.deployment_target  = '9.2'** (Got v9.2 from official Bugsnag pod spec https://github.com/CocoaPods/Specs/blob/master/Specs/c/1/5/Bugsnag/6.1.0/Bugsnag.podspec.json)
- changed **s.platform = :ios, '8.0'** to **s.ios.deployment_target  = '8.0'**
- changed **s.frameworks = 'MessageUI', 'SystemConfiguration'** to **s.framework = 'SystemConfiguration'**
- added **s.ios.framework = 'MessageUI'** (this is only compatible with iOS)

Full BugsnagReactNative.podspec file:
`require 'json'

Pod::Spec.new do |s|
  # NPM package specification
  package = JSON.parse(File.read(File.join(File.dirname(__FILE__), 'package.json')))

  s.name           = 'BugsnagReactNative'
  s.version        = package['version']
  s.license        = 'MIT'
  s.summary        = 'Bugsnag crash and error reporting for React Native apps'
  s.author         = { 'Delisa Mason' => 'iskanamagus@gmail.com' }
  s.homepage       = "https://docs.bugsnag.com/platforms/react-native"
  s.source         = { :git => 'https://github.com/bugsnag/bugsnag-react-native.git', :tag => "v#{s.version}"}
  s.ios.deployment_target  = '8.0'
  s.tvos.deployment_target  = '9.2'
  s.preserve_paths = '*.js'
  s.libraries      = 'z', 'c++'
  s.framework     = 'SystemConfiguration'
  s.ios.framework   = 'MessageUI'
  

[...]

PR #158, opened 2017-10-06
Add tvos example

fractalwrench:
Creates a (very basic) example for tvOS which notifies a non-fatal error. Additional setup steps required from the developer include:

- Opening their RN-generated xcodeproj, and adding `libz.tbd` and`libBugsnagReactNativeTvOs.a` to "Link binary with libraries" to the tvOS target
- Add `libReact.a` and remove version added by automatically generated RN project

To test this out will require replacing the contents of `node_modules/bugsnag_react_native/cocoa` with the `cocoa` folder in the project root.

See #152


fractalwrench:
N.B. There isn't currently a module.modulemap as this was causing a clash with the iOS target.

kattrali:
@fractalwrench what was the error message generated by having a modulemap?

fractalwrench:
@kattrali the error message was "Redefinition of module Bugsnag". There were two identical module maps (one in the iOS folder, the other in the tvOS folder). Is there any way to offer the same functionality without renaming the exported tvOS module?

fractalwrench:
@kattrali I've found a workaround for the modulemap issue by renaming the file and specifying its location in the xcodeproj Build Settings.

PR #159, opened 2017-10-06
Add option to leave breadcrumbs for console logs, fixes #82

### LINKED ISSUE #82: Is there a recommended way to overwrite console.log to send breadcrumbs to BugSnag?
I'm using the following code in my project but wanted to know if overwriting console.log was already built in or if there is a better way to overwrite console.log?

    //-------------------------------------------------------------------------------------------------
    // Create a singleton instance of the bugsnag client so we don't have to duplicate our configuration
    // anywhere.
    //-------------------------------------------------------------------------------------------------
    // https://docs.bugsnag.com/platforms/react-native/#basic-configuration
    import Config from '../data/Config';
    import { Client, Configuration } from "bugsnag-react-native";

    const configuration = new Configuration();
    configuration.apiKey = Config.BUGSNAG_API_KEY;
    configuration.notifyReleaseStages = ['production'];

    const client = new Client(configuration);

    (function(){
        var _log = console.log;

        console.log = function(logMessage){
            client.leaveBreadcrumb(logMessage, {type: 'log'});
            _log.apply(console,arguments);
        };

    })();
    //-------------------------------------------------------------------------------------------------
    export default {
        client
    };

bengourley:
With this PR, setting `config. consoleBreadcrumbsEnabled = true` or calling `client.enableConsoleBreadcrumbs()` will cause `console`'s logging functions to be wrapped. The arguments are passed to the original function and also sent to `client.leaveBreadcrumb()`.

I made an attempt to reasonably serialise most kinds of structures that get passed, but without adding another external dependency. I'm sure we could improve this in some way, but for now, the following output yields a dashboard state per the screenshot:

```js
console.log('jim')
console.debug('jim10')
console.info({ x: [1,2,3], y: 'yyy', z: new Date() })
console.warn('danger', 123)
console.error(new Error('foobar'))
```
🙃 _note the breadcrumbs render in reverse chronological order_
image


src/Bugsnag.js
```diff
@@ -3,6 +3,9 @@ import { NativeModules } from 'react-native';
 const NativeClient = NativeModules.BugsnagReactNative;
 
 const BREADCRUMB_MAX_LENGTH = 30;
+const CONSOLE_LOG_METHODS = [ 'log', 'debug', 'info', 'warn', 'error' ].filter(method =>
+  typeof console[method] === 'function'
```
bengourley: it appeared (in the android emulator at least) that `console.debug()` does not exist. However, it does in Chrome/Node and so rather than fully hardcode this list I thought I'd list all of the console methods that we would want to wrap, and filter that list by which actual exist in the current environment.

PR #165, opened 2017-10-11
Support setting `autoNotify` to disable native crash reporting

kattrali:


android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -217,6 +217,14 @@ private void configureRuntimeOptions(Client client, ReadableMap options) {
...
+          if (options.getBoolean("autoNotify")) {
+              client.enableExceptionHandler();
+          } else {
+              client.disableExceptionHandler();
+          }
+      }
+
       if (options.hasKey("codeBundleId")) {
           String codeBundleId = options.getString("codeBundleId");
           if (codeBundleId != null && codeBundleId.length() > 0)
```
fractalwrench: It would be nice to extract the literals in this file into a PSFS constant, as then if our JS API ever changed, we'd only need to edit the value in one place.

cocoa/BugsnagReactNative.m
```diff
@@ -230,7 +232,9 @@ + (void)startWithConfiguration:(BugsnagConfiguration *)config {
                             withValue:codeBundleId
                         toTabWithName:@"app"];
     }
-    if (![Bugsnag bugsnagStarted]) {
+    if ([Bugsnag bugsnagStarted] && !config.autoNotify) {
+        bsg_kscrash_setHandlingCrashTypes(BSG_KSCrashTypeUserReported);
+    } else if (![Bugsnag bugsnagStarted]) {
         [Bugsnag startBugsnagWithConfiguration:config];
```
fractalwrench: It appears that for this clause, fatal exceptions will still be reported even if `config.autoNotify` is false

android/src/main/java/com/bugsnag/BugsnagReactNative.java
kattrali: Yeah. Added a note for this for the backlog.

PR #170, opened 2017-10-23
Replace PropTypes from React with form prop-types package

sabbyt54:
See https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes
Also seems to be breaking on react@16.0.0-beta.5 and react-native@0.49.3


fractalwrench:
Thanks for the contribution @sabbyt54 - this all looks good to me.

PR #171, opened 2017-10-23
Add version bump script

fractalwrench:


fractalwrench:
@kattrali updated the changeset. I've kept the cocoapods gitignore as otherwise Pods could be added to VCS.

kattrali:
We can put `Pods` in the root .gitignore file to the same effect.

PR #182, opened 2017-11-14
Typescript definitions

bengourley:
Hey @cbrevik. I've merged your `.d.ts` file here and opened up a new PR as I added a bunch of QA/testing to it. Please could you install this branch and ensure the types work for you before we make a release?

Something along the lines of:

```
npm install "bugsnag/bugsnag-react-native#cbrevik-typescript-definition"
```

…should work! Let me know how you get on, and many thanks for the contribution 😃


index.d.ts
```diff
@@ -0,0 +1,92 @@
+export class Client {
+  public config: Configuration;
```
cbrevik: Unlike for example C#, all properties are public by default in TypeScript, so it's not necessary to mark everything as public. Not a usual practice since it makes the code more verbose than it needs to be.

index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+  ): void;
+
+  public clearBeforeSendCallbacks(): void;
+
+  public toJSON(): any;
+}
+
+type BeforeSend = (report: Report) => boolean | void;
+
+export class IStandardDelivery {
```
cbrevik: The class is named `StandardDelivery` in your library. Since the definition should mirror the implementation itself, it's not a good idea to rename it in the definition file. If someone tries to construct this it will fail at runtime. Also capital `I` before a type name is usually reserved for interfaces (though strictly not necessary).

index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+  public context?: string;
+  public groupingHash?: string;
+  public metadata: IMetadata;
+  public severity: "warning" | "error" | "info";
+  public stacktrace: string;
+  public user: any;
+
+  constructor(apiKey: string, error: Error);
+
+  public addIMetadata(
```
cbrevik: This function should not be renamed either, calling `addIMetdata` will fail at runtime. Also not usual practice to specify capital `I` for interfaces in functions that take them as parameters.

bengourley:
Hey again @cbrevik, thanks so much for spending the time reviewing this!

I've addressed 2/3 of your comments – the bits to do with "I"s being where they shouldn't. That was a case of over-zealous use of find and replace on my part 😉 

Regarding your other comment around the `public` declarations – the style I am using is the "recommended" ruleset of [tslint](https://palantir.github.io/tslint/). It will help us maintain the typedefs because it is something we can run at CI time, and catches a whole bunch of issues and errors that the typescript compiler will be permissive about. I don't want to switch off any rules unless there is a good reason to and I don't see any harm in leaving this one on. However if you think this is a _terrible_ idea, I'm happy to revert. You've clearly been using typsecript longer that I have and may have a better suggestion in this area.

Cheers!

cbrevik:
I usually set the `member-access` linting rule to `false` in my projects.

It seems like the idea behind that rule is that it  [(..) can make code more readable and accessible for those new to TS](https://palantir.github.io/tslint/rules/member-access/).

Which I can understand if you come from C# or Java, where you have to mark class members as `public` in order to expose them publicly. In those languages members are private by default though, so they'll probably either way be confused.

It's not a _terrible_ idea though, just not something I find a lot of projects actually do. And people coming from JavaScript will not be surprised I think. I have no strong opinions on the matter here though.

bengourley:
Hmm, interesting. That does give me a slight motivation to remove it. However, I'll leave it for now seeing as the path of least resistance is to leave it passing `tslint` in its current state 😁

Did you run this with your ts/bugsnag-react-native project? If so did it compile and work as expected? I'll crack on and merge if we're good in that respect.

cbrevik:
Yep, I tried it on an internal project here now, and it compiled just fine on the Bugsnag API surface we're using. 👍 

Looks good to me!

PR #183, opened 2017-11-14
Make console breadcrumbs safe to use

jamesreggio:
While the console breadcrumbs are great in concept, they've actually led to more crashes in our app than they've helped to diagnose. This PR resolves the underlying issue and adds safeguards to prevent other undiagnosed issues from crashing people's apps.

**The underlying issue**

If a `console.log` statement receives `null` or `undefined` as an argument, the code will throw a `TypeError` because you cannot call `toString()` on these values. In all cases, it's safer to use the `String` constructor than to call `toString()` on a value. (The former will ultimately delegate to `toString()` if defined, but will gracefully handle primitives like `null` and `undefined`.)

**The safeguard**

As new features are added to the tracker, it's important to wrap them in a top-most `try/catch`. We all make mistakes and cannot fully anticipate whether new code is going to result in problems. However, a top-level `try/catch` is an inexpensive way to mitigate the fallout from any unforeseen corner cases.

These precautions are especially important for a product like this.

### DIFF
--- a/src/Bugsnag.js
+++ b/src/Bugsnag.js
@@ -176,22 +176,26 @@ export class Client {
     CONSOLE_LOG_METHODS.forEach(method => {
       const originalFn = console[method];
       console[method] = (...args) => {
-        this.leaveBreadcrumb('Console', {
-          type: 'log',
-          severity: /^group/.test(method) ? 'log' : method,
-          message: args
-            .map(arg => {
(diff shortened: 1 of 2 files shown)


jamesreggio:
👍  Can you publish a patch release, too?

bengourley:
Yes, it's on its way 🚢

PR #185, opened 2017-11-16
automate vendored code update

fractalwrench:


Makefile
```diff
@@ -15,3 +15,14 @@ ifeq ($(VERSION),)
 endif
 	make VERSION=$(VERSION) bump && git commit -am "v$(VERSION)" && git tag v$(VERSION) \
 	&& git push origin && git push --tags && npm publish
+
+upgrade_vendor:
+ifeq ($(VERSION),)
+	@$(error VERSION is not defined. Run with `make VERSION=number upgrade_vendor`)
+endif
+	@sed -i '' "s/bugsnag-android:.*\'/bugsnag-android:$(VERSION)\'/" android/build.gradle
```
kattrali: I'd define a new variable for this. It wasn't clear that by running `make VERSION=foo bump upgrade_vendor` would set the android version to `foo`.

Makefile
```diff
@@ -15,3 +15,14 @@ ifeq ($(VERSION),)
...
+	cd ../bugsnag-cocoa
+	git checkout master && git pull
```
kattrali: haha having a slightly different directory structure saved me from doing something unfortunate here. Didn't realize it was going to wipe what I had in `bugsnag-cocoa`. Whew.

Probably should specify a defined release number for iOS too, to ensure that a tag is fetched rather than anything which might be unreleased. It'll keep the notifier version metrics correct and help with debugging.

fractalwrench:
@kattrali I've updated the logic to use a separate flag for android/ios version, and also to checkout the tagged version for iOS rather than master.

PR #187, opened 2017-11-17
Improve handled state validation

fractalwrench:
Improve validation of handled state to address 2 scenarios:

1) if the `notify` method was passed an object as its 5th parameter, `handledState` would not be created
2) if the `report._handledState` was modified in a `beforeSendCallback`, the `severityReason` could potentially be invalid

### DIFF
--- a/src/Bugsnag.js
+++ b/src/Bugsnag.js
@@ -329,8 +329,8 @@ export class Report {
   }
 
   toJSON = () => {
-    if (!_handledState || !(_handledState instanceof HandledState)) {
-      _handledState = new HandledState('warning', false, 'handledException');
+    if (!this._handledState || !(this._handledState instanceof HandledState)) {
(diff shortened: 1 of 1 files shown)


src/Bugsnag.js
```diff
@@ -309,7 +309,7 @@ export class Report {
     this.stacktrace = error.stack;
     this.user = {};
 
-    if (!_handledState) {
+    if (!_handledState instanceof HandledState) {
```
bengourley: You should write `!(_handledState instanceof HandledState)` because of [this reason](https://eslint.org/docs/rules/no-unsafe-negation).

Also we should have a linter on this project!

bengourley:
This seems ok, it's more checking than before, but there are still ways that rogue data could cause it to crash. I think it might be worth repeating the check on L312 in the toJSON method, since the entire `_handledState` object could have been replaced or meddled with.

PR #197, opened 2018-01-04
Add example build to CI

Cawllec:
Adds build stages to:
- Build an iOS example app
- Build an Android example app

As extra tests to ensure building correctly.


.travis.yml
```diff
@@ -1,7 +1,50 @@
...
+        - npm install
+        - ./node_modules/.bin/react-native link
+        - cd android
+        - ./gradlew assembleRelease
+      licenses:
+        - 
+    - stage: build cocoa
+      language: objective-c
+      osx-image: xcode7.3
+      env: SDK=macosx10.11 BUILD_OSX=1
```
fractalwrench: Is it possible to use a more modern version of XCode? Most people will be using 9 these days

.travis.yml
```diff
@@ -1,7 +1,50 @@
...
+      install: npm install
+      script: npm run test:unit:js && npm run test:smoke:typescript
+    - stage: build android
+      language: android
+      sudo: false
+      android:
+        components:
+         - tools
+         - platform-tools
+         - tools
```
fractalwrench: Is tools being listed twice intentional? There used to be a Travis bug that required a similar workaround, so this can be ignored if it's still an issue

Cawllec: I think this was necessary, although I can try without to double check

Cawllec:
I'm still getting a lot of warnings from xcodebuild, these can be suppressed by redirecting to /dev/null and requiring manual checking if the build ever fails.

Issue #203, opened 2018-01-20
[Typescript] Configuration object missing autoCaptureSessions property definition


raduflp:
In index.d.ts the newly added flag `autoCaptureSessions` is missing from the Configuration definition.

PR #209, opened 2018-01-25
Update typescript defs

fractalwrench:
Updates TypeScript definitions as types have changed since the [last edit](https://github.com/bugsnag/bugsnag-react-native/compare/v2.7.0...master). Also adds a note to the Contributing instructions to prevent this in future

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,6 +28,8 @@ export class Configuration {
   public codeBundleId?: string;
   public autoNotify: boolean;
   public handlePromiseRejections: boolean;
+  public autoCaptureSessions: boolean;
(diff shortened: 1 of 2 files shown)


fractalwrench:
#203

index.d.ts
```diff
@@ -41,15 +43,18 @@ export class Configuration {
 
   public clearBeforeSendCallbacks(): void;
 
+  public startSession(): void;
```
bengourley: This looks like it should be added to the `Client` class (not `Configuration`)

fractalwrench:
Updated to add the additional methods, and moved startSession to the Client.

PR #210, opened 2018-01-26
update android sdk versions

fractalwrench:
Updates gradlew for example apps, and moves to using API 27 SDK/build tools, and AGP 3 where possible.

The codepush example appears to require some window overlay permissions, so the targetSdk  has been left at 22 for now to avoid dynamically requesting those.


kattrali:
Merged as a part of #208

Issue #211, opened 2018-01-30
android build failed in 2.9.0


EthanChouTW:
### Steps to reproduce
1. install `bugsnag-react-native 2.9.0`
2. build with `react-native run-android`
3. build failed

```
/node_modules/bugsnag-react-native/android/src/main/java/com/bugsnag/BugsnagReactNative.java:265: error: method startFirstSession in class Client cannot be applied to given types;
          client.startFirstSession();
                ^
  required: Activity
  found: no arguments
  reason: actual and formal argument lists differ in length

:bugsnag-react-native:compileReleaseJavaWithJavac FAILED
```
### Expected behavior

build successful in 2.8.0 and before

### Observed behavior

should be recently changed, really fresh.
because of no activity sent in as argument to bugsnag android module
https://github.com/bugsnag/bugsnag-react-native/blob/master/android/src/main/java/com/bugsnag/BugsnagReactNative.java#L265

https://github.com/bugsnag/bugsnag-android/pull/231/files#diff-9775a55c8cf757a20a877341c819e7cfR327

### Environment

- bugsnag-react-native version: 2.9.0

fractalwrench:
Thanks for the report @EthanChouTW - this should be fixed in [v2.9.1](https://github.com/bugsnag/bugsnag-react-native/releases/tag/v2.9.1).

PR #212, opened 2018-01-30
Add missing param to client method

fractalwrench:
Addresses #211


Cawllec:
Is there a possibility that `getCurrentActivity` will return `null`?  Is this handled?

fractalwrench:
@Cawllec good catch - there's a chance that the activity could be null, I've updated to handle this case. I've also updated the iOS example app bundle id so that the app can be built

PR #222, opened 2018-03-12
fix(typedMap): Support converting circular/nested structures

### LINKED ISSUE #218: Maximum call stack size exceeded notifying Native iOS errors
### Steps to reproduce

1. Notify an error in an objective-c.

### Expected behavior
Bugsnag to successfully notify an error in an objective-c file.

### Observed behavior
Creates a new error by infinitely calling typedMap.
Stacktrace: 
```
RangeError Maximum call stack size exceeded. 
    [native code] includes
    node_modules/bugsnag-react-native/lib/Bugsnag.js:385:34 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
    ...
    node_modules/bugsnag-react-native/lib/Bugsnag.js:382:39 typedMap
```

### Environment

- react-native version: 0.51
- bugsnag-react-native version: 2.9.1
- react-native-code-push version (if any): N/A
- cocoapods version (if any): N/A
- iOS/Android version(s): iOS 11.2.6/11.2.5
- simulator/emulator or physical device?: Both
- debug mode or production?: Both

### Additional information
Sample code:
```
RCT_EXPORT_METHOD(updateAccessory:(NSString *)hwid property:(NSString *)property withValue:(NSString *)value
                  updateAccessoryWithResolver:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
  RCTLog(@"HWID: %@, PROPERTY: %@, VALUE: %@", hwid, property, value);
  [Bugsnag leaveBreadcrumbWithBlock:^(BugsnagBreadcrumb *crumb) {
    crumb.name = @"updateAccessory called";
    crumb.type = BSGBreadcrumbTypeLog;
    crumb.metadata = @{
                       @"hwid": hwid,
                       @"property": property,
                       @"value": value
                       };
  }];

...

[[self findSetpointCharacteristic:acc] writeValue:[NSNumber numberWithFloat: [value floatValue]] completionHandler:^(NSError *error) {
          if (error) {
            [Bugsnag notifyError:error];
            reject(@"Error", @"Error changing setpoint", error);
          } else {
            resolve(@"Successfully changed setpoint");
          }
        }];
        break;
      }

....

}
```

Trying to notify an error here is where it occurs the most. Commenting out all breadcrumbs in the trace still produces this issue.

bengourley:
typedMap(obj) would exceed the max stack size if it was given a circular or deeply nested data
structure. This change replaces circular references and levels of nesting above 10 with the strings
"[circular]" and "[max depth exceeded]" respectively.

Fixes #218.

---

Things to note/review:
- is it ok to use default arguments? It doesn't seem like they are used anywhere else
- is the horrible literal nesting ok in the tests? The other option was to generate the structure with code but I think it makes the logic of the test harder to follow if we do that
- are the substitute values ok? `{ type: 'string', value: '[<message>]' }` 
- I had to `export { typedMap }` in order to access it in the test, which means as a side effect it's available publicly. It might be better to pull it out into its own file?
- And while we're on the subject of that, it could maybe do with a more illustrative name? `convertObjectToTypedMap`, `serializeForNativeLayer`…


src/__tests__/Bugsnag.test.js
```diff
@@ -337,3 +337,138 @@ test('{enable|disable}ConsoleBreadCrumbs(): gracefully handles serialization edg
...
+  expect(typedMap(objA)).toEqual({ ref: { type: 'string', value: '[circular]' } })
+
+  const objB = {};
+  objB.leafA = {};
+  objB.leafB = {};
+  objB.leafA.ref = objB.leafB
+  objB.leafB.ref = objB.leafA
+
+  // this tests subtrees referring to eachother
+  expect(typedMap(objB)).toEqual({
```
Cawllec: Is the order in which the trees will be traversed guaranteed?  To avoid false negatives on this test.

bengourley: I don't think it's _guaranteed™_, but I'm almost certain that V8 will always traverse it like this.

src/Bugsnag.js
```diff
@@ -378,7 +378,8 @@ const allowedMapObjectTypes = ['string', 'number', 'boolean'];
  * Convert an object into a structure with types suitable for serializing
  * across to native code.
  */
-const typedMap = function(map) {
+const typedMap = function(map, maxDepth = 10, depth = 0, seen = new Set()) {
```
kattrali: Are default argument values created globally or at invocation time? In Python, I'd set seen to null and and create inside the body if it is null, as the default argument value is created once and any mutation would propagate.

bengourley: Invocation time: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Evaluated_at_call_time

PR #227, opened 2018-03-16
Add nativeserializer to packaging step

fractalwrench:


bengourley:
This fixes the current problem but it leaves us open to the same issue if/when we add more files within `src`.

I tried out the following command, which worked. I also used expanded argument names just to make things more obvious (I had to go digging to confirm what they meant)…

```
babel --source-maps --out-dir='lib' --ignore='__tests__' src
```

PR #240, opened 2018-05-15
fix(Errors): Enable nativeSerializer to handle Error objects

### LINKED ISSUE #239: Support for Error class in NativeSerializer (feature request)
### Description
Consider bugsnag code like below that tries to leverage metadata to track details of error.

```:JavaScript
const err = new Error('this is cause');
const params = { foo: 1 };
bugsnag.notify(new Error('This is an example'), (report) => {
  report.metadata.account = { cause: err, params };
});
```

It will create meatadata like:
```
{
  cause: {},
  params: { foo: 1 }
}
```

Note cause is logged as empty object.

**Describe the solution you'd like**
Implement handling for Error type within https://github.com/bugsnag/bugsnag-react-native/blob/master/src/NativeSerializer.js so it'll log relevant info (eg., error message)

Cawllec:
## Goal

Fixes #239, extracting the stack and message from a given `Error` class before serialising it.

## Design

Error objects do not have an iterator therefore are ignored by the `for` loop.  My initial thoughts were to add an iterator using `Object.keys` to extract all properties into a map, but the Error object does not respond to this either.

### Added

Added non-exported method `extractErrorDetails` to `NativeSerializer` that takes an error and returns an object with the Errors `message` and `stack` properties.

## Tests

Added a unit test to `NativeSerializer.test`

### Linked issues

Fixes #239 

  - [ ] Structural review of the classes, functions, and properties modified
  - [ ] Final review

- [x] Consistency across platforms for structures or concepts added or modified
- [x] Consistency between the changeset and the goal stated above
- [x] Internal consistency with the rest of the library - is there any overlap between existing interfaces and any which have been added?
- [x] Usage friction - is the proposed change in usage cumbersome or complicated?
- [x] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?
- [x] Concurrency concerns - if components are accessed asynchronously, what issues will arise
- [x] Thoroughness of added tests and any missing edge cases
- [x] Idiomatic use of the language

### DIFF
--- a/src/NativeSerializer.js
+++ b/src/NativeSerializer.js
@@ -1,3 +1,5 @@
+const isError = require('iserror')
+
 const allowedMapObjectTypes = [ 'string', 'number', 'boolean' ]
 
 /**
@@ -7,7 +9,7 @@ const allowedMapObjectTypes = [ 'string', 'number', 'boolean' ]
 const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()) => {
   seen.add(map)
   const output = {}
-  if (map instanceof Error) {
+  if (isError(map)) {
     map = extractErrorDetails(map)
   }
   for (const key in map) {
@@ -39,11 +41,8 @@ const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()
 }
 
 const extractErrorDetails = (err) => {
-  const {message, stack} = err
-  return {
-    message: message,
-    stack: stack
-  }
+  const { message, stack, name } = err
+  return { message, stack, name }
 }
 
-export default serializeForNativeLayer
+module.exports = serializeForNativeLayer
--- a/src/NativeSerializer.js
+++ b/src/NativeSerializer.js
@@ -7,6 +7,9 @@ const allowedMapObjectTypes = [ 'string', 'number', 'boolean' ]
 const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()) => {
   seen.add(map)
   const output = {}
+  if (map instanceof Error) {
+    map = extractErrorDetails(map)
+  }
   for (const key in map) {
     if (!{}.hasOwnProperty.call(map, key)) continue
 
@@ -33,6 +36,14 @@ const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()
     }
   }
   return output
+}
+
+const extractErrorDetails = (err) => {
+  const {message, stack} = err
+  return {
+    message: message,
+    stack: stack
+  }
 }
 
 export default serializeForNativeLayer
--- a/src/__tests__/NativeSerializer.test.js
+++ b/src/__tests__/NativeSerializer.test.js
@@ -144,4 +144,8 @@ test('serializeForNativeLayer handles error objects', () => {
   })
   expect(serialized['stack']['type']).toEqual('string')
   expect(serialized['stack']['value'].length).toBeGreaterThan(0)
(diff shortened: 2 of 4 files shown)


src/NativeSerializer.js
```diff
@@ -7,6 +7,9 @@ const allowedMapObjectTypes = [ 'string', 'number', 'boolean' ]
 const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()) => {
   seen.add(map)
   const output = {}
+  if (map instanceof Error) {
```
bengourley: I would use the [iserror](https://github.com/yefremov/iserror)  module for more robust `Error` detection. `instanceof` checks alone can yield [unexpected results](https://groups.google.com/forum/#!msg/comp.lang.javascript/XTWYCOwC96I/70rNoQ3L-xoJ). 

We make similar use of this module in [bugsnag-js](https://github.com/bugsnag/bugsnag-js/blob/9b2309d2a3adedb187dd908e7c4850d228ea2c46/browser/plugins/unhandled-rejection.js#L74).

src/NativeSerializer.js
```diff
@@ -35,4 +38,12 @@ const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()
   return output
 }
 
+const extractErrorDetails = (err) => {
+  const {message, stack} = err
```
bengourley: style nit that the linter doesn't pick up on:

```js
const { message, stack } = err
```

src/NativeSerializer.js
```diff
@@ -35,4 +38,12 @@ const serializeForNativeLayer = (map, maxDepth = 10, depth = 0, seen = new Set()
...
+  return {
+    message: message,
```
bengourley: there's nice new-ish js shortand for this:

```js
return { message, stack }
```

but you should also return the `name` and potentially the `code` property (this is only relevant to DOMexceptions, can you get them in RN? 🤔):

Cawllec: From what I've been reading the `message` and `stack` are the only guarantees.  I don't want to be dumping a whole load of `undefined`s into the metadata, so left out everything else.

I could add test whether properties are assigned and extract them if so?

src/__tests__/NativeSerializer.test.js
```diff
@@ -134,3 +134,14 @@ test('serializeForNativeLayer handles deep nesting', () => {
...
+test('serializeForNativeLayer handles error objects', () => {
+  const err = new Error('Oh no')
+  const serialized = serializeForNativeLayer(err)
+  expect(serialized['message']).toEqual({
+    type: 'string',
+    value: 'Oh no'
+  })
+  expect(serialized['stack']['type']).toEqual('string')
+  expect(serialized['stack']['value'].length).toBeGreaterThan(0)
+})
```
bengourley: add an assertion for `name` property, and if the `code` property is added, assertions for the inclusion/exclusion of that.

Cawllec:
Made changes based on the above review.  @fractalwrench, is there a way I can test that the `iserror` module is in-fact bundled with the module?

kattrali:
@Cawllec building the package via `yarn pack`/`npm pack` and then installing it into a sample app (something like `yarn add /path/to/bugsnag-react-native-2.9.4.tgz`) should both:

* Show that `iserror` is installed in the log output
* Be immediately testable

Cawllec:
After sorting out my local examples I can confirm the `iserror` module is being installed.  The errors attached to metadata are also coming through as expected when manually testing.

PR #246, opened 2018-06-01
refactor: Reduce gratuitous logging

### LINKED ISSUE #245: Allow Bugsnag's breadcrumb logger to be turned off (Android)
### Description
Superfluous logging output with breadcrumbs.

### Issue

The Bugsnag SDK outputs the breadcrumbs to Logcat before it saves the breadcrumb. This is flooding our console (and other log capturing SDK's ) with a ton of erroneous log statements. 

https://github.com/bugsnag/bugsnag-react-native/blob/master/android/src/main/java/com/bugsnag/BugsnagReactNative.java#L92

You should expose an API to disable this logger.

### Environment

Library versions: react native bugsnag v 2.9.3

kattrali:
Fixes #245

## Goal

Reduce the amount of logging that seems unnecessary even at a debug level


fractalwrench:
In addition to this change, I think we also should disable logging in production using the same method as the Android notifier.

kattrali:
@fractalwrench I'm not sure what you mean. Can you elaborate?

PR #251, opened 2018-06-21
Session tracking on by default

fractalwrench:
## Goal

Enable automatic session tracking by default in React Native.

## Changeset

- Updates Cocoa to the current state of the pre-release next branch, and Android to v4.5.0.
- Slight alterations to the logic for passing the notify/session endpoints to the native layer
- JS Configuration is now true by default

## Tests

One manual test on both Cocoa/Android to ensure that a session is sent by default.

### Outstanding Questions

- Will need updating to the released version of Cocoa.

## Review

  - [ ] Final review

### DIFF
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -225,7 +225,7 @@ public class BugsnagReactNative extends ReactContextBaseJavaModule {
       if (notify != null && notify.length() > 0) {
           config.setEndpoints(notify, sessions);
       } else if (sessions != null && sessions.length() > 0) {
-          BugsnagReactNative.logger.warn("The session tracking endpoint should not be set without the error reporting endpoint.")
+          BugsnagReactNative.logger.warn("The session tracking endpoint should not be set without the error reporting endpoint.");
       }
 
 
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -213,15 +213,14 @@ public class BugsnagReactNative extends ReactContextBaseJavaModule {
       }
 
       if (options.hasKey("endpoint")) {
-          String endpoint = options.getString("endpoint");
-          if (endpoint != null && endpoint.length() > 0)
-              client.setEndpoint(endpoint);
-      }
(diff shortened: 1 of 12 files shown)


fractalwrench:
One potential issue I can see is that we currently advise to initialise the notifier in the following ways:

```
[BugsnagReactNative start];
BugsnagReactNative.start(this);
```

This instantiates a bugsnag client immediately, which as part of construction, sends any stored sessions immediately to sessions.bugsnag.com by default. An alternative would be to document that this is the case, and advise that the existing `BugsnagReactNative.start(this, config);` option is used for both clients instead, in the case of on-prem.

Also, as noted in the description, this will require some more manual testing once the Cocoa notifier has been fully released.

android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -213,15 +213,14 @@ private void configureRuntimeOptions(Client client, ReadableMap options) {
...
-          String endpoint = options.getString("sessionsEndpoint");
-          if (endpoint != null && endpoint.length() > 0)
-              config.setSessionEndpoint(endpoint);
+          if (notify != null && notify.length() > 0) {
+              if (options.hasKey("sessionsEndpoint")) {
+                  String sessions = options.getString("sessionsEndpoint");
+                  config.setEndpoints(notify, sessions);
+              }
+          }
       }
```
kattrali: There should be a log message here if only `sessionsEndpoint` is set

kattrali:
> An alternative would be to document that this is the case, and advise that the existing `BugsnagReactNative.start(this, config);` option is used for both clients instead, in the case of on-prem.

Agreed, this should also happen in case there's a crash between native and JS initialization times.

android/src/main/java/com/bugsnag/BugsnagReactNative.java
fractalwrench: A warning is logged by bugsnag-android, if that's sufficient? https://github.com/bugsnag/bugsnag-android/blob/master/sdk/src/main/java/com/bugsnag/android/Configuration.java#L160

kattrali: In this case it would never get to the bugsnag-android warning since `setEndpoints` is only called if both are set.

PR #259, opened 2018-07-09
Address javac compiler warnings and intellij inspections

### LINKED ISSUE #250: BugsnagReactNative.java uses unchecked or unsafe operations
### Description
I followed native installation instructions for both iOS and Android. Works fine for iOS, but when building for Android via Android Studio I get the compile error below.

`.../BugsnagReactNative.java: uses unchecked or unsafe operations.`

Full error output is below, under _Error messages:_

### Environment

Library versions:

```shell
warning Resolution field "core-js@2.5.2" is incompatible with requested version "core-js@^1.0.0"
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ bugsnag-react-native@2.9.5
├─ react-native-code-push@5.3.2
└─ react-native@0.52.2
```

- Android version: 7.0 
- Physical device: Samsung Galaxy S7 Edge
- Debug mode
- `BugsnagReactNative.start(this)` is present in the
  `onCreate` method of my `MainApplication` class

**Specifying Google’s maven repository in android/build.gradle as:**
```
allprojects {
    repositories {
        maven { url "https://jitpack.io" }
        maven { url "https://mobile-sdk.jumio.com" }
        mavenLocal()
        jcenter()
        google()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url "$rootDir/../node_modules/react-native/android"
        }
        maven {
            url "https://maven.google.com"
        }
    }
}

```

**Error messages:**
Error when building from Android Studio
```
/Users/.../.../node_modules/bugsnag-react-native/android/src/main/java/com/bugsnag/BugsnagReactNative.java:
uses unchecked or unsafe operations.
Recompile with -Xlint:unchecked for details.
```

fractalwrench:
## Goal

Resolve any compiler warnings by addressing the root cause, or suppressing the warning where that is not feasible.

## Changeset

- Avoid use of static import
- Send correct number of arguments to `String.format`
- Add generic arguments to raw map variables
- Avoid redundant string array instantiation
- Add missing non-null annotation
- Add serialVersionUID to serialisable custom exception
- Suppress type erasure/3rd party warnings

### Linked issues

Fixes #250 

## Review

- [x] A changelog entry was added for the goal of this pull request
  - [x] Final review

### DIFF
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -369,7 +369,6 @@ class DiagnosticsCallback implements Callback {
         return output;
     }
 
-    @SuppressWarnings("unchecked") // ignore type erasure when casting Map
     @Override
     public void beforeNotify(Report report) {
         report.setNotifierName(NOTIFIER_NAME);
@@ -386,8 +385,11 @@ class DiagnosticsCallback implements Callback {
             MetaData reportMetadata = report.getError().getMetaData();
             for (String tab : metadata.keySet()) {
                 Object value = metadata.get(tab);
+
                 if (value instanceof Map) {
+                    @SuppressWarnings("unchecked") // ignore type erasure when casting Map
                     Map<String, Object> values = (Map<String, Object>) value;
+
                     for (String key : values.keySet()) {
                         reportMetadata.addToTab(tab, key, values.get(key));
                     }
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -364,6 +369,7 @@ Severity parseSeverity(String value) {
         return output;
     }
 
+    @SuppressWarnings("unchecked") // ignore type erasure when casting Map
```
kattrali: Can this be scoped a bit smaller, e.g.:

```java
@SuppressWarnings("unchecked")
Map<String, Object> values = (Map<String, Object>) value;
```

fractalwrench: Reduced the suppression to local variable scope.

PR #260, opened 2018-07-18
Fix possible release stage mismatch between sessions and error reports

kattrali:
## Goal

It is possible for the native release stage to not match the eventual actual release stage set on the JS side, and sessions could have a different release stage from subsequent errors. This change disables the native logic for automatically reporting the first session replacing it with a call to `startSession` from `startWithOptions`.

This change assumes that the React Native layer is the primary way to interact with the application, and when this is not the case, a different session tracking setup should be used instead of the automatic session detection provided (which should be added to the documentation for session tracking). 

This is also required to update to bugsnag-cocoa@5.16.2 as discussed in https://github.com/bugsnag/bugsnag-cocoa/pull/295#discussion_r202384805.

### Changed

* Updated bugsnag-cocoa to v5.16.2
* Disabled automatic session tracking in cocoa's `startWithConfiguration` and android's `init` methods
* Added an automatically tracked session in `startWithOptions` if automatic
  session tracking is enabled in JS

## Tests

Manually tested:

* Starting an application with the default release stage settings to confirm a single session is sent on iOS and Android
* Starting an application with a custom release stage set in JS to confirm a single session is sent on iOS and Android
* Causing handled and unhandled errors to ensure session counts and release stages match between sessions and error reports on iOS and Android
* Disabling automatic session tracking to ensure no sessions were sent at launch/reload
* Disabling automatic session tracking and manually sending a session to ensure a single session is sent
* Manually sending sessions with automatic session tracking enabled to ensure that both automatic and manual sessions are recorded

### Alternative Approaches

This implementation sends a session each time Bugsnag is initialized on the JS side, regardless of whether the native side has been running previously. My initial implementation would check whether the native side already has session tracking enabled and only send a session if it was was not natively enabled, however, this led to an inconsistent experience when soft reloading the JS components but not subsequently getting a new session sent. 

### Outstanding Questions

While this removes the possibility of sending a session for the wrong release stage, it raises the possibility of reporting an error with no session set at all. Any native error which occurs before JS-side initialization occurs will not have an assigned session. Is there a better strategy for mitigating this? We could potentially recommend setting release stage during native setup if using customized settings, but that may require information not available before the JS runtime is initialized.

### Linked issues

Related to bugsnag/bugsnag-cocoa#295

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
- [x] A changelog entry was added for the goal of this pull request
  - [x] Final review


android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -46,14 +46,23 @@ public static ReactPackage getPackage() {
...
+    // way to interact with the application should instead leverage startSession
+    // manually.
+    client.setAutoCaptureSessions(false);
+    return client;
   }
 
   public static Client startWithApiKey(Context context, String APIKey) {
-    return Bugsnag.init(context, APIKey);
+    Client client = Bugsnag.init(context, APIKey);
+    client.setAutoCaptureSessions(false);
```
fractalwrench: It would be good to consolidate calls to `client.setAutoCaptureSessions(false)` so that they only occur in one method.

fractalwrench:
This does raise a lot of wider questions about how we should set configuration values. For example, if a native error occurs before the JS bridge is fully initialised, the [user information](https://docs.bugsnag.com/platforms/react-native/configuration-options/#setuser) may not be set. This also applies to several other values (e.g. releaseStage) - unless I'm mistaken errors can still theoretically be sent to the wrong releaseStage with this changeset?

I think the best way to _guarantee_ against these scenarios happening would be to recommend the following:

- JS-specific configuration options, such as `consoleBreadcrumbsEnabled` should be set within the JS layer.
- All other configuration options should be set within the native layer.
- Passing a `Configuration` instance parameter to `bugsnag.init` should be preferred, and not mutating fields after initialisation unless absolutely necessary (this is a general point that could apply to all platforms).

The downside of this is that it requires some additional configuration on the user's part, and requires separating configuration into two different places (although this may be useful depending on usecase). If achieving 100% consistency in native error configuration isn't a high priority, then that may make this changeset a reasonable approach.

@martin308 are any of the above points applicable to how the new C# layer in Unity configures the Android/Cocoa clients?

kattrali:
> This does raise a lot of wider questions about how we should set configuration values. For example, if a native error occurs before the JS bridge is fully initialised, the user information may not be set. This also applies to several other values (e.g. releaseStage) - unless I'm mistaken errors can still theoretically be sent to the wrong releaseStage with this changeset?

Correct, native errors which occur before React Native is initialized can have the wrong release stage, no user information, and no session. Default detection is still in place, so we would know if it was a debug build vs production, as is native version information. Its a similar situation to how server-side platforms work, where there is not user information nor a session for errors outside of request scope. I'm open to a better approach, though this one seems the least likely to result in confusion about illogical unhandled errors percentages where error counts are greater than session counts.

> I think the best way to guarantee against these scenarios happening would be to recommend the following:
> * ...
> * All other configuration options should be set within the native layer.

The problem here is that user and release stage information may not be available before the JS runtime is initialized. For example, build information could come from a package.json file or code push environment info. User information could be set via [rn-instagram-login](https://www.npmjs.com/package/rn-instagram-login) or similar.

kattrali:
> are any of the above points applicable to how the new C# layer in Unity configures the Android/Cocoa clients?

On the configuration side this isn't an issue, as the new client configuration basically disables everything and handles session tracking and handled exception reporting in the C# layer. 

Error reporting starts until Unity is initialized, so there's no intermediate state where partially configured errors are sent. (@martin308 feel free to chime in on this where I've missed something)

android/src/main/java/com/bugsnag/BugsnagReactNative.java
fractalwrench: I didn't notice the difference between config/client at first - this should be fine as-is, as long as it remains only 1 line of configuration.

PR #271, opened 2018-08-13
Update CodePush example app

fractalwrench:
Updates the codepush example app by regenerating the project with RN 0.56.

Note: the release instructions in the README currently seem to assume that only iOS should be built, I wasn't too sure on whether we need to support both, or whether we could get rid of this script entirely in favour of linking to the docs. I've left as-is for now.


examples/codepush/app/lib/bugsnag.js
```diff
@@ -3,7 +3,10 @@
 // anywhere.
 //-------------------------------------------------------------------------------------------------
 // https://docs.bugsnag.com/platforms/react-native/#basic-configuration
-import { Client } from 'bugsnag-react-native';
-const client = new Client();
+import { Client, Configuration } from 'bugsnag-react-native';
+
+const config = new Configuration('YOUR_API_KEY_HERE');
+config.codeBundleId = '1.0.0-b12'
```
martin308: is setting this important to the example? If so we should document why this is being done

examples/codepush/README.md
```diff
@@ -25,6 +25,12 @@ A sample application demonstrating using Bugsnag with React Native and CodePush.
...
 
-2. Run the application, deploying new bundles as needed using
+2. Add your Bugsnag API key in `bugsnag.js`:
+
+    ```js
+    const config = new Configuration('YOUR_API_KEY_HERE');
+    ```
+
+3. Run the application, deploying new bundles as needed using
    `npm run release -- [version number]`. Feel free to customize and reuse the
```
martin308: This seems to require `BUGSNAG_API_KEY` env variable to be set which is not documented

examples/codepush/app/lib/bugsnag.js
fractalwrench: Yes - I've added a link to the relevant docs.

fractalwrench:
@martin308 sorry about the long wait in replying to this. It'd be good to get your opinion on whether you think the `npm release` step is still helpful, which is why I haven't updated any of the steps to use AppCenter, or documented any required ENVARs yet. My current thinking was that we could change the README to add something like the following steps instead:

3. Deploy a new code bundle (linking to AppCenter instructions, adding a code sample, making a note about codeBundleId)

4. Upload source maps to Bugsnag (linking to our docs, adding a code sample)

The disadvantage is that this approach would be a bit more manual than `npm release`. However, it would mean our instructions work on Windows + Linux, and we wouldn't necessarily have to document setting various ENVARs. What are your thoughts?

PR #272, opened 2018-08-15
Update release notes to add network throttle steps

fractalwrench:


snmaynard:
Surely we should just change the endpoints to point somewhere that doesnt exist?

bengourley:
You'd get a quick connection fail in that case, right? But the point of this check (I believe) is to see if a slow request occupies the UI/main thread and makes the app seem unresponsive.

snmaynard:
https://httpstat.us/ use that?

fractalwrench:
The point of this step is to simulate a slow network connection to ensure that no ANRs/watchdogs are observed.

Attempting to reach a non-existent endpoint would test that the client fails gracefully and caches a request if the server isn't available. I can add that in as an additional step if you feel that's something we need to test manually on each release.

snmaynard:
I would prefer it was deterministic. Thats why putting a delay on https://httpstat.us/ is preferable in my mind.

fractalwrench:
@snmaynard I've updated the instructions to use httpstat.us, if you're happy with the approach + wording I'll update the other PRs.

snmaynard:
That seems good to me

PR #273, opened 2018-09-21
Remove Gradle warning

bahaa96:
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -23,6 +23,6 @@ android {
 }
 
 dependencies {
-    compile 'com.bugsnag:bugsnag-android:4.5.0'
-    compile 'com.facebook.react:react-native:+'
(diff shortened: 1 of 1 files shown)


fractalwrench:
Thanks for the contribution @bahaa96 - I believe we'd need to specify `com.bugsnag.android` as an `api` dependency as we would want to expose it transitively. It also looks like CI is failing on this PR - I think this would be down to our version of the gradle wrapper or maybe due to using an older version of the Android Gradle Plugin.

We do also have plans to address this internally if you didn't feel like updating this PR further.

fractalwrench:
Closing as we plan to address this internally and the PR fails on CI.

PR #278, opened 2018-10-01
[Typings] Allow undefined apiKeyOrConfig

danilobuerger:
`apiKeyOrConfig === undefined` is being handled at https://github.com/bugsnag/bugsnag-react-native/blob/master/src/Bugsnag.js#L21

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,7 @@
 export class Client {
   public config: Configuration;
 
-  constructor(apiKeyOrConfig: string | Configuration);
+  constructor(apiKeyOrConfig?: string | Configuration);
 
   public notify(
     error: Error,


fractalwrench:
Thanks for opening a PR @danilobuerger - what would the use-case be for passing `undefined` into the constructor?

snmaynard:
You shouldn't pass undefined in here so the types are correct. I don't think the fact that we cope with undefined being passed in when you aren't using a type system imply that the types should also allow undefined.

danilobuerger:
@snmaynard so why does the code explicitly handle this case (as I already linked):
```
if (typeof apiKeyOrConfig === 'string' || typeof apiKeyOrConfig === 'undefined') {
  this.config = new Configuration(apiKeyOrConfig)
}
```
If I shouldn't pass `undefined` why the explicit check for it here? Why allow it then and create a working `Configuration`?

danilobuerger:
Also why handle it in the native module (`undefined` in js is `null` in java):
```
          if (apiKey != null) {
              client = Bugsnag.init(this.reactContext, apiKey);
          } else {
              client = Bugsnag.init(this.reactContext);
          }
```

danilobuerger:
There are a lot more samples I could post here that show that a `undefined` api key is actually properly being handled. And if that isn't enough, here the use case:

> Your Integration API Key, available on your Bugsnag dashboard under “Settings”. If not set, it will be loaded from your AndroidManifest.xml or Info.plist if available on Android and iOS respectively.

By passing `undefined` here (which again even the Docs say is allowed), the api key will be loaded from the manifest of plist.

PR #279, opened 2018-10-02
[Typings] setUser optional params

danilobuerger:
The params to `setUser` should be optional to allow submitting `null` values as coded here:
https://github.com/bugsnag/bugsnag-react-native/blob/master/android/src/main/java/com/bugsnag/BugsnagReactNative.java#L152
https://github.com/bugsnag/bugsnag-react-native/blob/master/cocoa/BugsnagReactNative.m#L192


danilobuerger:
@bengourley I don't really care about `null`. If you want it in, I will amend the PR.

danilobuerger:
Ah, I only meant the native side. Sorry for the confusion.

fractalwrench:
Merged in #301 - thanks for the contribution!

PR #285, opened 2018-10-31
Update plain example app to latest version of react native

fractalwrench:
Updates the React Native plain example app to use the latest version of React Native. The current example app does not build successfully when using XCode 10 due to a [variety](https://github.com/facebook/react-native/issues/14382) [of](https://github.com/facebook/react-native/issues/19569) [issues](https://github.com/facebook/react-native/issues/15432).

This was achieved by running `react-native-git-upgrade` and resolving conflicts manually. I have tested this on the Android + iOS apps running in release mode, and confirmed that errors are reported for each available button.

We may also want to create tickets to update the cocoapods and codepush sample projects, as they will be affected by the same issue.


snmaynard:
Can we get CI passing?

fractalwrench:
I've been attempting to get CI passing on a separate test branch - a bit of additional config is required to get Travis working with XCode 10.

fractalwrench:
I believe this should now pass CI, by changing to an xcode 10 image on Travis and using the legacy build system. See this issue for further detail: https://github.com/facebook/react-native/issues/19573

PR #290, opened 2018-11-20
fix(iOS): Ensure reports on disk prior to app termination

kattrali:
## Goal

(Extends #287)

Ensure report delivery while holding the native JS execution thread for as little time as possible in the case of non-fatal errors being reported. Updated the integration test to make the stack trace ridiculous.

## Tests

* Tested against physical devices as well as in the integration suite

### Linked issues

Closes #287


cocoa/BugsnagReactNative.m
```diff
@@ -141,7 +141,9 @@ + (void)startWithConfiguration:(BugsnagConfiguration *)config {
 
 RCT_EXPORT_MODULE()
 
-RCT_EXPORT_METHOD(notify:(NSDictionary *)options) {
+RCT_EXPORT_METHOD(notify:(NSDictionary *)options
+                  resolve:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject) {
```
fractalwrench: We should update the declaration in the header file also

android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -107,13 +107,9 @@ public void leaveBreadcrumb(ReadableMap options) {
...
   }
 
-  @ReactMethod
-  public void notify(ReadableMap payload) {
-      notifyBlocking(payload, false, null);
-  }
 
   @ReactMethod
-  public void notifyBlocking(ReadableMap payload, boolean blocking, com.facebook.react.bridge.Callback callback) {
+  public void notify(ReadableMap payload, com.facebook.react.bridge.Promise promise) {
```
fractalwrench: We could get away without the fully qualified class name here - it was necessary previously to disambiguate between the two `Callback` classes

kattrali:
Both nice catches, thanks @fractalwrench !

PR #299, opened 2018-12-18
feat: Whitelist files to be included in npm package

bengourley:
Without this whitelist, arbitrary things from the publishing developer's filesystem can be
unintentionally included in the tarball that gets uploaded to npm. The "files" property in
package.json is used to specify the only files/directories that should be included.

I tested this locally by running `npm pack`. There are quite a lot of things in the `android` and `cocoa` dirs, so to check for the presence of interesting files you can do:

`npm pack` followed by `tar -tvf bugsnag-react-native-*.tgz | grep lib/`


fractalwrench:
There are a couple of files in the `android` directory which could be filtered out:

`local.properties`
`android.iml`

Otherwise I'm happy with these changes provided CI passes.

bengourley:
@fractalwrench How's this update? 18fdee8 

You can't really ignore files using `files` in package.json, so I added them to the `.npmignore` in line with to the following [npm docs](https://docs.npmjs.com/files/package.json#files)…

> You can also provide a .npmignore file in the root of your package or in subdirectories, which will keep files from being included. At the root of your package it will not override the “files” field, but in subdirectories it will. The .npmignore file works just like a .gitignore. If there is a .gitignore file, and .npmignore is missing, .gitignore’s contents will be used instead.

Notably I didn't have these files on my local filesystem, so I didn't see them get correctly ignored.

PR #303, opened 2019-01-08
Replace gradle "compile" with "implementation" configuration

pqkluan:
## Goal

To remove this warning:
> Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
> It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

### Changed

Replace all the occurs of "compile" with "implementation" configuration in "./android/build.gradle" as the warning advise.

## Tests

Nothing changes

### Alternative Approaches

There are none.

### Outstanding Questions

None.

### Linked issues

Not found.


alizbazar:
What's the status with this?

fractalwrench:
@alizbazar we have an internal ticket to track this. As `implementation` and `api` are not present in older versions of Gradle, we plan on waiting until most users are likely to be using v3+ of the Android Gradle Plugin before merging this change.

nofxincubus:
What is the exact timeline for this? "We plan on waiting until most users are likely to be using v3+" seems like a very vague timeline for us to depend on.

duncanhewett:
@nofxincubus `compile` is removed in Gradle 5 which is not yet the default in the latest React Native version. What version of React Native, Gradle and Android Gradle Plugin are you using?

This issue is that using `implementation` will break compatibility with the default Android Gradle Plugin version (2.x) used in React Native <0.56 (Mid 2018).

nofxincubus:
We are using RN 0.57.2, Gradle 4.8,  Gradle Plugin 3.2.1. Should we fork for now until update's available?

fractalwrench:
@nofxincubus you should be able to continue to use bugsnag-react-native with that version of Gradle, but will get a warning. Just to confirm, are you encountering any roadblocks to building your application when Bugsnag is integrated, or are you simply observing this warning?

nofxincubus:
We see it as warning now. We have every other packages updated except for this one so just wanted to know the ETA.

faketon:
`api 'com.bugsnag:bugsnag-android:4.11.0'`
`implementation 'com.facebook.react:react-native:+'`

I test with above code and work well but below code have error when build in my app.

`implementation 'com.bugsnag:bugsnag-android:4.11.0'`
`implementation 'com.facebook.react:react-native:+'`

timothystewart6:
It won't compile with `implementation` as @faketon  commented it needs to be `api`, at least in its current state.

fractalwrench:
Thanks for reporting this issue - I've created a PR to address this in #362, so am closing this.

PR #306, opened 2019-01-17
Update examples to show how to deobfuscate JS + Native stacktraces

fractalwrench:
## Goal

The example apps don't currently show how to deobfuscate JS stacktraces by uploading source maps, or how to deobfuscate Android/iOS stacktraces by following the steps on the native platform. Adding a working example of how to do this should make it easier for users to understand how to integrate Bugsnag into their projects.

## Changeset

- Added the bugsnag-android-gradle plugin, which automatically uploads ProGuard/NDK files
- Added a script with the necessary steps to upload dSYMs, and an example dSYMs zip.
- Added scripts which run all the necessary steps to upload sourcemaps, for Android/iOS, on debug/release builds.
- Updated the README to give a high-level overview of how deobfuscation works and linked to additional documentation on our website.


examples/plain/ios/upload-react-native-dsyms.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+bugsnag-dsym-upload example-dsyms.zip
```
bullmo: Rather than uploading an example dSYM zip it could upload the dsym generated for the example app

e.g.
```
bugsnag-dsym-upload build/Build/Products/Release-iphonesimulator/
```

examples/plain/scripts/upload-debug-android.sh
```diff
@@ -0,0 +1,16 @@
...
+
+API_KEY=YOUR_API_KEY_HERE # set your own API key here
+APP_VERSION=1 # set your own app version here (using the Android versionCode)
+
+# Download debug source maps from Metro bundler
+curl "http://localhost:8081/index.bundle?platform=android&dev=true&minify=false" > android-debug.bundle
+curl "http://localhost:8081/index.bundle.map?platform=android&dev=true&minify=false" > android-debug.bundle.map
+
+# Upload source maps to Bugsnag, making sure to specify the correct app-version.
+bugsnag-sourcemaps upload \
```
bullmo: `bugsnag-sourcemaps` needs to be installed for this to work. Maybe we should have a comment to indicate that.

fractalwrench:
Thanks for the feedback @bullmo - I've added a note on pre-requisite installations in the README, and addressed the other feedback inline.

snmaynard:
Yeah you can `yarn run` too

examples/plain/scripts/upload-debug-android.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
```
bengourley: you should put [`set -e`](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script) somewhere near the top of these scripts, so that if any of the commands fail, it stops and outputs the error.

When I ran this it told me the `.map` file was invalid because even though the `curl` command failed, it still wrote the output to the `.map` file. While true, this was not the source of the error so is misleading.

examples/plain/package.json
```diff
@@ -4,7 +4,13 @@
...
-    "test": "jest"
+    "test": "jest",
+    "upload-sourcemaps:debug:android": "./scripts/upload-debug-android.sh",
+    "upload-sourcemaps:release:android": "./scripts/upload-release-android.sh",
+    "upload-proguard:release:android": "./android/upload-proguard.sh",
+    "upload-sourcemaps:debug:ios": "./scripts/upload-debug-ios.sh",
+    "upload-sourcemaps:release:ios": "./scripts/upload-release-ios.sh",
+    "upload-dsyms:release:ios": "./ios/upload-react-native-dsyms.sh"
   },
   "dependencies": {
```
bengourley: You should install bugsnag-sourcemaps in this package so that these commands "just" work.

This should do the trick:

```
yarn add --dev bugsnag-sourcemaps
```

examples/plain/README.md
```diff
@@ -52,3 +52,45 @@ A robust example of how to make best use of the Bugsnag React Native notifier.
...
    react-native run-android --variant=release
    ```
+
+## Deobfuscating stacktraces
+
+This app provides working examples of how to deobfuscate stacktraces for JS, Android, and iOS. For further information, please consult [our React Native documentation](https://docs.bugsnag.com/platforms/react-native/showing-full-stacktraces/).
+
+### JavaScript
+
+You will need to upload source maps to Bugsnag in order to deobfuscate JavaScript stack traces. Example scripts for achieving this on Android/iOS in both debug and release builds can be found [here](scripts). The example script requires the installation of [bugsnag-sourcemaps](https://github.com/bugsnag/bugsnag-sourcemaps) to work. The scripts can also be invoked in this example project using [npm run](https://docs.npmjs.com/cli/run-script.html):
```
bengourley: Rather than say it requires the installation of `bugsnag-sourcemaps`, add it as a dev dependency to the project. See other comment.

fractalwrench: I've updated the wording here slightly to state that it's installed in this example project as a dev dependency.

PR #317, opened 2019-03-05
Add mazerunner scenarios for stopping and resuming sessions

fractalwrench:
## Goal

Verifies that when `stopSession` and `resumeSession` are called, the session information in the error payload is updated on both iOS + Android.

## Changeset

- Added an additional mazerunner step that allows the scenario to be specified by a string
- Added scenario where `stopSession` is called and no session info is present in the error payload
- Added scenario where `resumeSession` is called and session info is present in the error payload
- Exposed `bugsnag` client so that it can be used in the fixture

Note: these scenarios do not currently run on CI. I have tested them locally and they pass on my machine.


features/session_stopping.feature
```diff
@@ -0,0 +1,26 @@
...
+    And the request 0 is valid for the session tracking API
+    And the request 1 is valid for the error reporting API
+    And the payload field "events.0.session" is null for request 1
+
+    Examples:
+    | platform |
+    | Android  |
+    | iOS      |
+
+Scenario Outline: When a session is resumed the error uses the previous session
```
bengourley: It doesn't seem to me that the assertions in this test are doing what the outline says. I would expect to see an assertions that given the following sequence of events:

```
- startSession()
- send report
- stopSession()
- send report
- resumeSession()
- send report
```

The following conditions are met:

```
- report 1 and report 3 have the same session id
- report 2 has no session
- there was one session received
- report 1 has one report in its session
- report 3 has two events in its session
```

features/fixtures/sampler/index.js
```diff
@@ -8,6 +8,8 @@ config.delivery = new StandardDelivery(endpoint, endpoint)
 config.autoCaptureSessions = false
 const _bugsnag = new Client(config)
 
+export default _bugsnag;
```
bengourley: Why are all these named with underscores? In JS a `_` prefix is used to denote "private" APIs by convention (where it can't be enforced).

Unless there is a reason for this, I think this identifier should just have the `_` removed wherever it is used.

features/steps/app_steps.rb
```diff
@@ -9,6 +9,9 @@ def launch_crashy_app(os, event)
   run_required_commands([["features/scripts/launch_#{os}_app_crash.sh"]])
 end
 
+When(/^I launch an (\w+) app with "(\w+)"$/) do |os, event_type|
```
bengourley: I would be nice to convert the previous test cases to use this generic step…

a) for consistency
b) so that not every new test has to write a new step
c) the steps below can all be removed

It could also have a more descriptive title than "with", e.g.:

`"When I run the {string} scenario on a(n?) {string} device"`

features/fixtures/sampler/index.js
fractalwrench: Good to know - it was named like this before when it was only used in this file. I'll update the identifier

PR #320, opened 2019-03-11
fix: BSGConnectivity hostName

Jekiwijaya:
## Goal
Fixing `https://notify.bugsnag.com` is being unresolved, since it's include the protocol, the accepted one is only hostname.
Group

https://developer.apple.com/documentation/systemconfiguration/1514904-scnetworkreachabilitycreatewithn
> nodename
The node name of the desired **host**. This name is the same as that passed to the gethostbyname or getaddrinfo functions. The value of this parameter is copied into the new object.

## Changeset
Changes
```diff
- NSString *hostName = [url absoluteString];
+ NSString *hostName = [url host];
```

## Tests
It should be resolved
Group2

## Review

  - [ ] Initial review of the intended approach, not yet feature complete
  - [x] Final review


Jekiwijaya:
https://github.com/bugsnag/bugsnag-cocoa/pull/332

lightswitch05:
Thank you for fixing this. Over the past 7 days there have been 10 failed requests on my network due to this bug. Its quite annoying to see full URLs popping up in my DNS records

image

PR #327, opened 2019-03-24
Remove support library dependency to support RN 0.59

kdawgwilk:
This fixes issues for those upgrading to RN 0.59 and also makes the process easier for those migrating to androidx

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -24,6 +24,6 @@ android {
 }
 
 dependencies {
-    compile 'com.bugsnag:bugsnag-android:4.12.0'
-    compile 'com.facebook.react:react-native:+'
+    implementation 'com.bugsnag:bugsnag-android:4.12.0'
+    implementation 'com.facebook.react:react-native:+'
 }
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -2,7 +2,6 @@ package com.bugsnag;
 
 import android.app.Activity;
 import android.content.Context;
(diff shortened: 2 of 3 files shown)


Cawllec:
Hi @kdawgwilk, thanks for the pull request.  Looking over the changes made here I'm not sure that the issues you're experiencing relate to our notifier.  However I can raise some of the issues here so they get properly addressed.

Specifically for some of the changes:
- The `compile` -> `implementation` gradle deprecation is a warning we'll address and ensure is dealt with in a backwards compatible way, but shouldn't be effecting react-native 0.59
- Likewise the `prepublish` -> `prepare` should be dealt with in a similar manner, but shouldn't effect RN 0.59
- The Android support library and corresponding annotation shouldn't be an issue in RN 0.59 as the default `build.gradle` for the RN android project contains the support library by default:
`implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"`
as previous versions of RN have.  Have you modified the build.gradle in anyway to perhaps remove this dependency?

I have tested our notifier in a react-native 0.59 app, and it launched without any problems, but if you continue to have issues it would be very useful if you could provide a reproduction case that we could use to figure out what's going on.

lucasbento:
I can confirm that this is not breaking on `0.59` as it's just a warning.

I also came here thinking that this was the reason as when it was failing in my project the first line would mention the `compile > implementation` deprecation, the reason I went for this dependency first is that it's literally the first one showing in my Android Studio.

fractalwrench:
The compilation issues with Androidx have been addressed in #361, so I'm closing this PR for now.

PR #331, opened 2019-04-04
Tests/anr detection maze tests

Cawllec:
## Goal
Adds test targeting the Android-only ANR detection feature.

Run locally, these tests have successfully passed.


features/fixtures/sampler/ios/sampler.xcodeproj/project.pbxproj
```diff
@@ -40,6 +39,8 @@
 		832341BD1AAA6AB300B99B32 /* libRCTText.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 832341B51AAA6A8300B99B32 /* libRCTText.a */; };
 		ADBDB9381DFEBF1600ED6528 /* libRCTBlob.a in Frameworks */ = {isa = PBXBuildFile; fileRef = ADBDB9271DFEBF0700ED6528 /* libRCTBlob.a */; };
 		F7023905233C48118ADFE9D6 /* libz.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 41820A385C704024802F33EE /* libz.tbd */; };
+		16C36C00A7E041AAB76EF567 /* libReactNativeConfig.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 873456FFA9624C3CA83A37C3 /* libReactNativeConfig.a */; };
```
fractalwrench: Why are there ios changes when this PR adds android tests?

features/support/env.rb
```diff
@@ -20,13 +20,24 @@ def launch_packager
   run_script('features/scripts/launch_packager.sh')
 end
 
+def set_envfile(envfile)
+  Dir.chdir('features/fixtures/sampler') do
+    FileUtils.rm('.env')
+    FileUtils.cp("scenario_envs/#{envfile}.env", '.env')
```
fractalwrench: We should add error handling to all uses of FileUtils so that an error is logged if mazerunner fails to perform a file operation - otherwise it'll be hard to debug scenarios.

features/fixtures/sampler/scenario_envs/default-anr.env
```diff
@@ -0,0 +1,2 @@
+MAZE_ENABLE_ANR=true
+MAZE_ANR_TIMEOUT=0
```
fractalwrench: I think this needs a comment to state that values <= 0 won't be set - this was confusing to start with

features/fixtures/sampler/android/app/src/main/java/com/sampler/MainApplication.java
```diff
@@ -43,6 +49,21 @@ public ReactNativeHost getReactNativeHost() {
...
   public void onCreate() {
     super.onCreate();
+    boolean enableANR = false;
+    int anrTimeout = 0;
+    try {
+      ApplicationInfo ai = getPackageManager().getApplicationInfo(getPackageName(), PackageManager.GET_META_DATA);
+      String rawEnableANR = (String) ai.metaData.get("enable-anr");
+      enableANR = Boolean.parseBoolean(rawEnableANR);
+      anrTimeout = Integer.parseInt((String) ai.metaData.get("anr-timeout"));
+    } catch (NameNotFoundException e) {
```
fractalwrench: We should rethrow this so that the error will be recorded in the maze_output directory, as this will make debugging scenarios much easier.

features/fixtures/sampler/android/app/src/main/java/com/sampler/CustomMethodsModule.java
```diff
@@ -0,0 +1,43 @@
...
+import com.facebook.react.bridge.NativeModule;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.bridge.ReactMethod;
+import com.bugsnag.android.Bugsnag;
+import com.bugsnag.android.Client;
+import com.bugsnag.android.Configuration;
+
+public class CustomMethodsModule extends ReactContextBaseJavaModule {
```
fractalwrench: I think longer-term it'd be good to create a separate module for each thing we need to test, i.e. one for ANR, one for session tracking, etc. Otherwise this class could gradually gain more and more responsibilities over time and become large and unwieldy. Maybe we'd want to rename this module something related to ANRs.

features/fixtures/sampler/android/app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,12 @@
       android:icon="@mipmap/ic_launcher"
       android:allowBackup="false"
       android:theme="@style/AppTheme">
+      <meta-data
+        android:name="enable-anr"
+        android:value="@string/MAZE_ENABLE_ANR"/>
```
fractalwrench: How does that work? There isn't anything in strings.xml and from looking only at the code, I don't really understand where the value is coming from.

Cawllec: Ok, I'm using https://github.com/luggit/react-native-config to pull values from a .env into the environment at build time.  They are available in various places, including JS, Gradle, and as strings in the Manifest.  https://github.com/luggit/react-native-config#android

features/fixtures/sampler/ios/sampler.xcodeproj/project.pbxproj
Cawllec: This is due to the linking of the React-Native-Config module, as mentioned below.  I think it's worth linking it to both flavours, as if we come to use it again it'd be strange that it's only linked to one.

PR #336, opened 2019-04-15
Run checkstyle and lint on Android code, address existing violations

fractalwrench:
## Goal

Code in bugsnag-react-native did not run checkstyle or Android Lint automatically, which would have helped detect code smells and generally improve code quality. Running these tools on CI should catch issues early and prevent us from shipping violations.

## Changeset

- Setup Android Lint to abort the build on any errors not included within a baseline of ignored issues, in the example project and Android code
- Setup Checkstyle to abort the build on any style errors in the example project and Android code
- Pre-existing violations were also addressed, including:

  - Splitting top-level classes into separate files
  - Using an indent of 4 rather than 2
  - Adding missing braces to conditionals
  - Adding default branch to switch statements
  - Ordering imports
  - Adding missing override modifiers
  - Fixing modifier ordering

## Tests

- Ran the new inspections both locally and on CI
- Ran mazerunner scenarios for Android

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ jobs:
         - yarn install
         - ./node_modules/.bin/react-native link
         - cd android
-        - ./gradlew assembleRelease
+        - ./gradlew lint checkstyle assembleRelease
       licenses:
         -
     - stage: build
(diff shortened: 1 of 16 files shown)


fractalwrench:
Thanks @Pezzah - I've resolved the remaining comment.

PR #340, opened 2019-04-23
Specify right dir for public headers to fix CocoaPods

kdawgwilk:
## Goal

Build fails when using CocoaPods because there are duplicate headers being added by CocoaPods because the directory was not specific enough.

Screen Shot 2019-04-23 at 11 46 12 AM

## Tests

Tested locally editing my `BugsnagReactNative.podspec` file and running a `pod install` and then building


fractalwrench:
Thanks for raising this PR @kdawgwilk - we've released v2.17.1 which does not contain these build files. We're going to tighten up our package.json whitelist to ignore these sorts of files entirely, which will also fix the issue for everyone who doesn't use Cocoapods.

PR #345, opened 2019-04-25
Collect version information in device.runtimeVersions

fractalwrench:
## Goal
We should collect the React Native version in `device.runtimeVersions.reactNative` for all error reports and sessions.

Note: this PR relies on the following [Android](https://github.com/bugsnag/bugsnag-android/pull/474) and [Cocoa](https://github.com/bugsnag/bugsnag-cocoa/pull/341) PRs. The Cocoa changes have been vendored in; the Android changes require installing an artefact to mavenLocal in the usual way.

## Changeset
- Updated bugsnag-cocoa to `runtime-versions-hooks` branch
- Added `beforeSendSession` callback to both Cocoa/Android which adds the react native version to the session payload
- Added a `beforeSend` block/callback which adds the react native version to the error payload
- Calculated the React Native version by concatenating the values present in the map supplied by the React Native platform

## Discussion
There are a couple of important things to note with this approach:

1. We use a `beforeSend` callback on React Native rather than a `beforeNotify` callback because this is the only callback mechanism that is guaranteed to be invoked when a fatal error occurs in the NDK.

2. Ideally we would specify callbacks in a configuration parameter that is passed into a Bugsnag constructor; however, this is not possible with the current architecture of bugsnag-react-native. Instead we specify callbacks after constructing Bugsnag, which means there is a chance that some sessions could be missing runtime version information. This is a known issue with React Native configuration in general, and is something that we should address as part of a general revamp of the notifier.

## Tests
Manually tested that the field was present in the payload for an error/session on both Android/Cocoa.

Mazerunner scenarios will be added in a separate PR to aid review.

### DIFF
--- a/cocoa/BugsnagReactNative.m
+++ b/cocoa/BugsnagReactNative.m
@@ -267,12 +267,17 @@ RCT_EXPORT_METHOD(startWithOptions:(NSDictionary *)options) {
     config.automaticallyCollectBreadcrumbs = [RCTConvert BOOL:options[@"automaticallyCollectBreadcrumbs"]];
 
     [config addBeforeSendSession:^void(NSMutableDictionary *_Nonnull data) {
-        data[@"device"] = [self addDeviceRuntimeVersion:data[@"device"]];
+        data[@"device"] = [self addDeviceRuntimeVersion:data[@"device"]
+                                     reactNativeVersion:[self findReactNativeVersion]];
     }];
 
     [config addBeforeSendBlock:^bool(NSDictionary *_Nonnull rawEventData,
                                      BugsnagCrashReport *_Nonnull report) {
-        report.device = [self addDeviceRuntimeVersion:report.device];
+        NSString *reactNativeVersion = report.metaData[@"_bugsnag"][@"reactNativeVersion"];
+        if (reactNativeVersion != nil) {
+            report.device = [self addDeviceRuntimeVersion:report.device reactNativeVersion:reactNativeVersion];
+        }
+        report.metaData = [self removeRuntimeVersionFromMetaData:report];
 
         return !([report.errorClass hasPrefix:@"RCTFatalException"]
                  && [report.errorMessage hasPrefix:@"Unhandled JS Exception"]);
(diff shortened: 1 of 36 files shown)


cocoa/BugsnagReactNative.m
```diff
@@ -265,8 +266,14 @@ + (void)startWithConfiguration:(BugsnagConfiguration *)config {
     config.shouldAutoCaptureSessions = [RCTConvert BOOL:options[@"autoCaptureSessions"]];
     config.automaticallyCollectBreadcrumbs = [RCTConvert BOOL:options[@"automaticallyCollectBreadcrumbs"]];
 
+    [config addBeforeSendSession:^void(NSMutableDictionary *_Nonnull data) {
+        data[@"device"] = [self addDeviceRuntimeVersion:data[@"device"]];
+    }];
+
     [config addBeforeSendBlock:^bool(NSDictionary *_Nonnull rawEventData,
                                      BugsnagCrashReport *_Nonnull report) {
+        report.device = [self addDeviceRuntimeVersion:report.device];
```
kattrali: This will append the runtime version of the app when sending the report, which isn't necessarily the same as when capturing the crash.

cocoa/BugsnagReactNative.m
```diff
@@ -296,6 +308,63 @@ + (void)startWithConfiguration:(BugsnagConfiguration *)config {
...
+        runtimeVersions = [NSMutableDictionary new];
+    }
+    runtimeVersions[@"reactNative"] = reactNativeVersion;
+    copy[@"runtimeVersions"] = runtimeVersions;
+    return copy;
+}
+
+// see https://github.com/facebook/react-native/blob/6df2edeb2a33d529e4b13a5b6767f300d08aeb0a/scripts/bump-oss-version.js
+- (NSString *)findReactNativeVersion {
+    NSDictionary *versionMap = RCTGetReactNativeVersion();
```
kattrali: This call should be wrapped so it only has to happen once and is cached after that.

```objc
- (NSString *)findReactNativeVersion {
    static dispatch_once_t onceToken;
    static NSString *BSGReactNativeVersion = nil;
    dispatch_once(&onceToken, ^{
        NSDictionary *versionMap = RCTGetReactNativeVersion();
        NSNumber *major = versionMap[@"major"];
        NSNumber *minor = versionMap[@"minor"];
        NSNumber *patch = versionMap[@"patch"];
        NSString *prerelease = versionMap[@"prerelease"];
        NSMutableString *versionString = [NSMutableString new];

        if (![major isEqual:[NSNull null]]) {
            [versionString appendString:[major stringValue]];
            [versionString appendString:@"."];
        }
        if (![minor isEqual:[NSNull null]]) {
            [versionString appendString:[minor stringValue]];
            [versionString appendString:@"."];
        }
        if (![patch isEqual:[NSNull null]]) {
            [versionString appendString:[patch stringValue]];
        }
        if (![prerelease isEqual:[NSNull null]]) {
            [versionString appendString:@"-"];
            [versionString appendString:prerelease];
        }
        BSGReactNativeVersion = [NSString stringWithString:versionString];
    });
    return BSGReactNativeVersion;
}
```

Ref: [`dispatch_once`](https://developer.apple.com/documentation/dispatch/1447169-dispatch_once)

fractalwrench:
I've updated bugsnag-android and bugsnag-cocoa to the latest release, and also cached the React Native version on iOS as requested.

PR #346, opened 2019-04-29
fix: Make setUser() more robust

### LINKED ISSUE #343: bugsnag.setUser() number crashes
### Description
calling bugsnag.setUser() with an ID which is a number, causes an error.

### Issue
Screenshot 2019-04-24 at 14 26 43

### Environment
iOS Simulator

### Example code snippet

bugsnag.setUser(3, 'Jessica Jones', 'jess@example.com');

I beleive it should handle numbers too for IDs, at least it should not crash. At the very least the documentation should say that a string is required or it will crash.

bengourley:
The native implementation of `setUser()` only accepts strings, but it was possible to pass it other types of value from the JS layer. This PR updates the JS implementation to:

- attempt to stringify all values
- tolerate errors in stringification
- fallback to undefined when a string cannot be obtained

Fixes #343.

### Tests

As well as the unit tests added in this PR, this was tested manually on an iOS simulator with different values for `user.id` (`string`, `number`, `null`, `Object.create(null)`.


src/__tests__/Bugsnag.test.js
```diff
@@ -376,3 +376,22 @@ test('{enable|disable}ConsoleBreadCrumbs(): gracefully handles serialization edg
...
+  const config = new Configuration('API_KEY')
+  const c = new Client(config)
+  c.setUser('123', 'Jim', 'jim@bugsnag.com')
+  c.setUser(4, 'Jim', 'jim@bugsnag.com')
+  c.setUser(null, 'Jim', 'jim@bugsnag.com')
+  c.setUser(Object.create(null), 'Jim', 'jim@bugsnag.com')
+  expect(mockSetUser).toHaveBeenCalledTimes(4)
+  expect(mockSetUser.mock.calls[0][0]).toEqual({ id: '123', name: 'Jim', email: 'jim@bugsnag.com' })
+  expect(mockSetUser.mock.calls[1][0]).toEqual({ id: '4', name: 'Jim', email: 'jim@bugsnag.com' })
+  expect(mockSetUser.mock.calls[2][0]).toEqual({ id: 'null', name: 'Jim', email: 'jim@bugsnag.com' })
```
fractalwrench: Null is coerced into a string here as otherwise the device ID would be used as the user ID.

PR #348, opened 2019-05-02
Runtime-versions move into `device`: iOS & Native tests

Cawllec:
## Goal

Adds iOS tests, and native code tests targeting both Android and iOS.

## Tests

Passed locally

### DIFF
--- a/features/runtime_versions.feature
+++ b/features/runtime_versions.feature
@@ -1,6 +1,6 @@
 Feature: Runtime versions are included in all requests
 
-    Scenario Outline: Uncaught exception contains runtime versions for Android
+    Scenario Outline: Uncaught exception contains runtime versions
         When I launch an <platform> app which has an uncaught exception
         And I wait for 1 second
         Then I should receive a request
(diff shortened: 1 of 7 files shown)


features/runtime_versions.feature
```diff
@@ -2,24 +2,42 @@ Feature: Runtime versions are included in all requests
 
     Scenario Outline: Uncaught exception contains runtime versions for Android
         When I launch an <platform> app which has an uncaught exception
+        And I wait for 1 second
```
fractalwrench: Is there any way we can avoid this?

features/runtime_versions.feature
```diff
@@ -2,24 +2,42 @@ Feature: Runtime versions are included in all requests
 
     Scenario Outline: Uncaught exception contains runtime versions for Android
```
fractalwrench: We'll want to rename all the scenarios that refer to Android as they now run on both Android/iOS

features/runtime_versions.feature
Cawllec: Not from what I can tell, with no wait it fails the iOS tests.  However these should be removed fairly promptly as this repo is updated to MazeRunner 2.  So not ideal, but temporary

snmaynard: With maze 2 we can

PR #363, opened 2019-05-29
Remove unnecessary gradlew from android directory

fractalwrench:
## Goal

Remove unnecessary files from android directory, reducing artefact size. This was originally included for local development purposes and is no longer required.

## Changeset

Removed the gradle wrapper, which is unused.

## Tests

Ran existing CI suite.


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+## TBD
+
+* Remove unnecessary gradlew from android directory
```
kattrali: Should note here how this improves the package for users, reducing bundle size etc.

bengourley:
What about the file android `./gradlew`? Should that exist? If so should it have execute permissions? I remember trying to run it before and it not being executable.

PR #364, opened 2019-05-29
v2.20.0

fractalwrench:


android/build.gradle
```diff
@@ -23,8 +23,8 @@ android {
 }
 
 dependencies {
-    compile 'com.bugsnag:bugsnag-android:4.14.2'
-    compile 'com.facebook.react:react-native:+'
+    api 'com.bugsnag:bugsnag-android:4.14.2'
+    implementation 'com.facebook.react:react-native:+'
```
bengourley: Can you explain to me why one of these is `api` and the other is `implementation`? I trust that it was made in good judgement, I just want to understand it!

fractalwrench: Sure - `api` exports transitive dependencies to consumers of a module; `implementation` doesn't.

In the case of `bugsnag-react-native`, we want to expose `bugsnag-android` as part of our API, as we use types defined within that package in `BugsnagReactNative.java`. In the case of facebook's react-native package, we don't need to export it as a transitive dependency as the user will always have their own dependency setup in a React Native app.

The Android developers site has some useful docs on `api` vs `implementation`, and also some other different dependency configurations: https://developer.android.com/studio/build/dependencies#dependency_configurations

PR #380, opened 2019-07-16
Release v2.22.1

kattrali:


fractalwrench:
This should be a minor bump as it alters the native dependencies

kattrali:
It changes the native dependencies but doesn't alter the native interface.

PR #393, opened 2019-08-14
fix: make JavaScriptException inherit from BugsnagException

fractalwrench:
## Goal

As part of https://github.com/bugsnag/bugsnag-android/pull/553 the way exceptions are serialised in bugsnag-android was altered. This was to allow the exception name/message to be altered after a `BugsnagException` object has been constructed.

As part of this change, the exception serialisation was altered so that when building an error, all `Throwable` objects must be wrapped in a `BugsnagException` if they are [not already of that type](https://github.com/bugsnag/bugsnag-android/pull/553/files#diff-04f6e695a46ee00c79f74a90c284bb41R59). This meant that `JavaScriptException#toStream()` was not invoked when serialising an error report, leading to the name being serialised as `com.bugsnag.JavaScriptException`, rather than the name of the JS error.

This changeset alters `JavaScriptException` to inherit `BugsnagException`, meaning that it is not wrapped in a `BugsnagException`, and that its custom `toStream` implementation will be invoked and the JS error serialised correctly.

## Tests

Tested manually in the example app and confirmed that `ReferenceError` was the error class and the correct stacktrace was collected.

Additional unit test coverage has also been added in the Android PR.

### DIFF
--- a/android/src/main/java/com/bugsnag/BugsnagReactNative.java
+++ b/android/src/main/java/com/bugsnag/BugsnagReactNative.java
@@ -1,10 +1,12 @@
 package com.bugsnag;
 
+import com.bugsnag.BugsnagReactNative;
 import com.bugsnag.android.BreadcrumbType;
 import com.bugsnag.android.Bugsnag;
 import com.bugsnag.android.Client;
 import com.bugsnag.android.Configuration;
 import com.bugsnag.android.InternalHooks;
+import com.bugsnag.android.JavaScriptException;
 
 import android.content.Context;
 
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/bugsnag/BugsnagReactNative.java
```diff
@@ -27,7 +29,7 @@
     private ReactContext reactContext;
     private String libraryVersion;
     private String bugsnagAndroidVersion;
-    static final Logger logger = Logger.getLogger("bugsnag-react-native");
+    public static final Logger logger = Logger.getLogger("bugsnag-react-native");
```
bengourley: why is this change necessary?

fractalwrench: `JavaScriptException` needs to be in a different package to access package-private methods in `BugsnagException`, which meant it could no longer access this field.

PR #394, opened 2019-08-14
Hermes parsing

bengourley:
## Goal

Add support for [Hermes](https://facebook.github.io/react-native/docs/hermes) stacktraces.

## Changeset

The stacktrace parsing logic on Android is based in `JavaScriptException.java` so the changes are limited to that file. Hermes does not affect iOS, so no changes are required there.

A regex is used to detect what format of stacktrace we have before descending into the appropriate parsing routine.

## Tests

There were no unit tests for the `BugsnagException` class and the existing parsing routine so we added them. Four permutations of tests are now included:

- JSCore / dev
- JSCore / release
- Hermes / dev
- Hermes / release

I also manually tested these changes by installing the notifier on an example app and sending errors to a dashboard.

### Alternative Approaches

We discussed passing `global.HermesInternal` through to the native layer somehow but I couldn't figure out a watertight way to do this.

### Outstanding Questions

Source maps generated by Hermes live in a unique location and don't have sources included, so instructions to upload source maps in our docs are not adequate for projects running Hermes.


android/src/main/java/com/bugsnag/JavaScriptException.java
```diff
@@ -30,47 +30,108 @@ public void toStream(JsonStream writer) throws IOException {
...
-            writer.beginObject();
-            String[] methodComponents = rawFrame.split("@", 2);
-            String fragment = methodComponents[0];
-            if (methodComponents.length == 2) {
-                writer.name("method").value(methodComponents[0]);
-                fragment = methodComponents[1];
+
+        boolean isHermes = rawStacktrace.matches("(?s).*\\sat .* \\(.*\\d+:\\d+\\)\\s.*");
+        for (String frame : rawStacktrace.split("\\n")) {
+            if (isHermes) {
```
fractalwrench: Is there a scenario where a single stacktrace could contain frames from both JSCore and Hermes? If not, it'd be good to move the conditional to outside the for loop just to clarify this

android/src/main/java/com/bugsnag/JavaScriptException.java
```diff
@@ -30,47 +30,108 @@ public void toStream(JsonStream writer) throws IOException {
...
+        int methodEnd = frame.indexOf(" (");
+        boolean hasMethodInfo = methodStart < methodEnd;
+
+        // serialise srcInfo
+        if (hasSrcInfo || hasMethodInfo) {
+            writer.beginObject();
+            writer.name("method").value(frame.substring(methodStart, methodEnd));
+            if (hasSrcInfo) {
+                String srcInfo = frame.substring(srcInfoStart + 1, srcInfoEnd);
+                String file = srcInfo.replaceFirst(":\\d+:\\d+$", "");
```
fractalwrench: There are a couple of regexes in this class that are hardcoded as strings - it'd be good to extract them to a `private static final` constant and possibly add a quick comment summarising what they do

android/src/main/java/com/bugsnag/JavaScriptException.java
```diff
@@ -30,47 +30,108 @@ public void toStream(JsonStream writer) throws IOException {
...
+
+                writer.name("file").value(file);
+
+                String[] chunks = srcInfo.split(":");
+                Integer lineNumber = parseIntSafe(chunks[chunks.length - 2]);
```
fractalwrench: We should check the size of `chunks.length` here, just to be defensive about passing in a negative index into an array

android/src/main/java/com/bugsnag/JavaScriptException.java
bengourley: the stacktrace format is detected once for the entire thing – it's assumed that it can only contain one kind of frame. If you move the conditional out of the loop though, you have to write the loop twice?

fractalwrench: Completely misread this - originally thought `isHermes` was initialised inside the loop.

android/src/main/java/com/bugsnag/JavaScriptException.java
bengourley: I would agree if the regexes were used more than once, but to move them away from the place they're used would be less clear in my opinion. I'm happy to name and comment them, but I'd be against moving the definitions away from the location they're used, if that makes sense?

Is this recommendation purely a stylistic preference or are you suggesting the performance would be improved by making it static too?

fractalwrench: Adding comments would resolve this IMO - as they are only used in one place a constant might be overkill.

PR #395, opened 2019-08-16
Next release

fractalwrench:


tomlongridge:
Are we missing a reference to Hermes (#394) in changelog? Suggest that it needs to reference the docs and/or the need to use source map tool.

PR #400, opened 2019-09-03
Log isFatal to reported error

usrbowe:
## Goal

When logging JS errors, it's hard to see if app actually crashed. Or just some JS error was thrown, but user was not affected. 
Also if we use ErrorBoundary from React, we can handle lot of component errors, so app still keeps working.
This helps to understand, if app was crashed, and to prioritise fix on the error. 

### Added
Log `isFatal` via `report.metadata`

## Tests

Crash app with any unhandled JS error: `throw new Error('crash app')`

### Outstanding Questions

Is the `isFatal` reliable indicator? 

## Review

For the submitter, initial self-review:

- [ ] Commented on code changes inline explain the reasoning behind the approach
- [ ] Reviewed the test cases added for completeness and possible points for discussion
- [ ] A changelog entry was added for the goal of this pull request
- [ ] Check the scope of the changeset - is everything in the diff required for the pull request?
- This pull request is ready for:
  - [ ] Initial review of the intended approach, not yet feature complete
  - [ ] Structural review of the classes, functions, and properties modified
  - [ ] Final review

For the pull request reviewer(s), this changeset has been reviewed for:

- [ ] Consistency across platforms for structures or concepts added or modified
- [ ] Consistency between the changeset and the goal stated above
- [ ] Internal consistency with the rest of the library - is there any overlap between existing interfaces and any which have been added?
- [ ] Usage friction - is the proposed change in usage cumbersome or complicated?
- [ ] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?
- [ ] Concurrency concerns - if components are accessed asynchronously, what issues will arise
- [ ] Thoroughness of added tests and any missing edge cases
- [ ] Idiomatic use of the language

PR #402, opened 2019-09-10
feat: Include isFatal in unhandled error reports

bengourley:
Re-implement #400 without adding `lib` to source control and using a convenience method for updating metadata, to ensure it merges rather than overwrites.

Tested locally:
image


johnkiely1:
Closing as library has moved to bugsnag-js

PR #412, opened 2019-10-03
Expose OOM config to javascript

maxkomarychev:
Native library has two options to control OOM reporting but they are not
used in javascript side.
https://docs.bugsnag.com/platforms/ios/configuration-options/#reportbackgroundooms
https://docs.bugsnag.com/platforms/ios/configuration-options/#reportooms

## Goal

Allow OOM-related configurations from js configuration.

## Review

  - [ ] Initial review of the intended approach, not yet feature complete
  - [ ] Structural review of the classes, functions, and properties modified
  - [ ] Final review


mattdyoung:
Hi @maxkomarychev 

Thanks for the PR, but I'm afraid we won't be able to merge it at this time.

We're in the process of making some significant improvements to bugsnag-react-native, including how options like this are initialized. The next major release will use a different approach to initialization so this problem won't exist.

For the time being we'd recommend following our instructions here for enhanced native integration which will allow you to configure the underlying iOS options:
https://docs.bugsnag.com/platforms/react-native/react-native/enhanced-native-integration/#ios-native-installation

PR #423, opened 2019-11-07
Use h-char import for React/RCTVersion

matthova:
When manually linking bugsnag-react-native
```
#import "RCTVersion.h"
```

XCode throws the following error:
```
bugsnag-react-native/cocoa/BugsnagReactNative.m:4:9: 'RCTVersion.h' file not found
```

By instead using
```
#import <React/RCTConvert.h>
```
this build error is resolved

## Goal
Fix build error

## Changeset
Described above

## Tests
Tested by building for iOS

## Review

- [x] Commented on code changes inline explain the reasoning behind the approach
  - [ ] Final review

### DIFF
--- a/cocoa/BugsnagReactNative.m
+++ b/cocoa/BugsnagReactNative.m
@@ -1,7 +1,7 @@
 #import "Bugsnag.h"
 #import "BSG_KSCrashC.h"
 #import "BugsnagReactNative.h"
-#import "RCTVersion.h"
+#import <React/RCTVersion.h>
 #import <React/RCTConvert.h>
 
 NSString *const BSGInfoPlistKey = @"BugsnagAPIKey";


mattdyoung:
Hi @matthova 

I don't think this change shouldn't be necessary.

Can you share which versions of bugsnag-react-native and React Native itself you're using when you encounter this issue?

matthova:
Using
    "bugsnag-react-native": "2.23.2",
	react-native 0.59

mattdyoung:
@matthova 
I've been trying to reproduce the issue you describe on a clean project on React Native 0.59.x, but it seems to work for me.

Steps I've tried:
```
$ react-native init example --version react-native@0.59.10

$ cd example

$ npm install bugsnag-react-native --save

$ react-native link bugsnag-react-native
```

Added the following lines to `index.js`:
```
import { Client } from 'bugsnag-react-native';
const bugsnag = new Client('YOUR_API_KEY_HERE');
```
```
$ react-native run-ios
```

This builds and launches in an emulator without issue and also runs fine in XCode.

I've also gone on to integrate with the native layer too again without seeing the issue you describe:
https://docs.bugsnag.com/platforms/react-native/react-native/enhanced-native-integration/#ios-native-installation

Can you share steps to reproduce the issue which prompted you to raise this PR?

matthova:
@mattdyoung thanks for checking it out. I believe the issue is related to our Podfile's setup

Near the beginning we're declaring
```ruby
install! 'cocoapods',
  :generate_multiple_pod_projects => true,
  :incremental_installation => true
```

We were able to resolve this github issue by adding the following to postinstall
```ruby
post_install do |installer|
  installer.generated_projects.each do |project|
    project.targets.each do |target|
      if target.name == "BugsnagReactNative"
        puts "Adding 'React-Core/React' to HEADER_SEARCH_PATHS for #{target}"
        append_header_search_path(target, '${PODS_ROOT}/Headers/Public/React-Core/React/')
      end
    end
  end
end
```

Issue #433, opened 2019-12-13
v2.23.3 does not compile for iOS


antoinerousseau:
### Description

I upgraded `bugsnag-react-native` from 2.23.2 to 2.23.3 with Yarn, and ran `pod install` in `ios/`, but it broke the build.

### Issue
> ❌  error: node_modules/bugsnag-react-native/cocoa/vendor/bugsnag-cocoa/Configurations/Config.xcconfig: unable to open file (in project "Bugsnag") (in target 'BugsnagStatic' from project 'Bugsnag')

### Environment

Library versions:

```shell
yarn list v1.19.2
├─ bugsnag-react-native@2.23.2
└─ react-native@0.59.10
```

- cocoapods version (if any) (`pod --version`): 1.8.4
- iOS/Android version(s): iOS, N/A
- simulator/emulator or physical device?: N/A
- debug mode or production?: production

- [ ] (iOS only) `[BugsnagReactNative start]` is present in the
  `application:didFinishLaunchingWithOptions:` method in your `AppDelegate`
  class? no
- [ ] (Android only) `BugsnagReactNative.start(this)` is present in the
  `onCreate` method of your `MainApplication` class? no

phillipsam:
Hi @antoinerousseau - does this issue persist if you delete your node modules directory and reinstall?

antoinerousseau:
@phillipsam just tried it and still fails

```
[13:10:27]: $ set -o pipefail && xcodebuild -workspace ./ios/artefac.xcworkspace -scheme artefac -destination 'generic/platform=iOS' -archivePath /Users/antoine/Library/Developer/Xcode/Archives/2019-12-17/artefac\ 2019-12-17\ 13.10.27.xcarchive archive | tee /Users/antoine/Library/Logs/gym/artefac-artefac.log | xcpretty
[13:10:32]: ▸ ❌  error: /Users/antoine/Sites/artefac/mobile/node_modules/bugsnag-react-native/cocoa/vendor/bugsnag-cocoa/Configurations/Config.xcconfig: unable to open file (in project "Bugsnag") (in target 'BugsnagStatic' from project 'Bugsnag')
[13:10:32]: ▸ ❌  error: /Users/antoine/Sites/artefac/mobile/node_modules/bugsnag-react-native/cocoa/vendor/bugsnag-cocoa/Configurations/Config.xcconfig: unable to open file (in project "Bugsnag") (in target 'BugsnagStatic' from project 'Bugsnag')
[...]

PR #440, opened 2020-01-13
Test against freshly built projects on 0.59 and 0.61 on CI

kattrali:
This change updates the CI build strategy to create new React Native projects for RN 0.59 and RN 0.61 and check that a linked project compiles. This removes an existing test which added the bugsnag-react-native package to an existing 0.53 project. This change should ensure newer and older versions of RN build correctly with bugsnag-react-native and that issues like #433 are caught automatically.

~This is a draft awaiting #441 being merged, which will make the build pass.~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ android/local.properties
 android/android.iml
 test/tmp
 .gradle
+tmp-projects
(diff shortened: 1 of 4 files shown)


kattrali:
> Was this considered?

Actually yeah! The current supported minimum version is 0.40, which I was working on adding but then totally lost the thread when I picked up other things. Going to reinstate 0.53 in its place.

PR #455, opened 2020-04-28
TARGET_OS_IPHONE used when not defined

jerrymarino:
The way that this define works is that TargetConditionals.h uses checks
that are set implcitly by clang.

Before using TARGET_OS_* TargetConditionals.h should be imported or it
won't work. Often in bugsnag this is implicitly imported by Foundation.h

### DIFF
--- a/cocoa/vendor/bugsnag-cocoa/Source/BSGOutOfMemoryWatchdog.m
+++ b/cocoa/vendor/bugsnag-cocoa/Source/BSGOutOfMemoryWatchdog.m
@@ -1,3 +1,4 @@
+#import <TargetConditionals.h>
 #if (TARGET_OS_TV || TARGET_OS_IPHONE)
 #define BSGOOMAvailable 1
 #else
--- a/cocoa/vendor/bugsnag-cocoa/Source/BugsnagCrashReport.m
+++ b/cocoa/vendor/bugsnag-cocoa/Source/BugsnagCrashReport.m
@@ -6,6 +6,7 @@
 //
 //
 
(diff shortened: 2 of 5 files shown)


bolsinga:
alternately these .m files could import their .h first, and the read the `TARGET_OS_*` stuff.

bolsinga:
It may be worth noting that this issue would be found if the Xcode build used no prefix headers or precompiled headers.

jerrymarino:
@bolsinga where are you seeing a prefix header used with bugsnag - I didn't see one in this repo

bolsinga:
> @bolsinga where are you seeing a prefix header used? I didn't notice one in the repo or podspec. That being said, I noticed a few places that `TargetConditionals.h` wasn't imported, where this define is defined. Simply put, the code needs to import `TargetConditonals.h` in some shape or form before using this define

This code doesn't import "TargetConditionals.h" anywhere manually, so the only way I understand that the .m files affected would compile "in the wild" is with some sort of a prefix header afaik.

mattdyoung:
We believe this issue has now been fixed in https://github.com/bugsnag/bugsnag-cocoa/pull/654, released in [v6.0.0](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.0.0).

Note: There are a number of breaking changes in v6.0.0 as part of an effort to improve the interface and make it more consistent across our error reporting libraries, so please refer to the upgrade guide:
https://github.com/bugsnag/bugsnag-cocoa/blob/master/UPGRADING.md

